PR #3, opened 2018-04-05
Implement Simple Naive Bayesian filter

vmax:


projects/noloan/app/filter/bayes/BayesFilter.java
```diff
@@ -0,0 +1,168 @@
...
+        WordInfo wordInfo = message.get(i);
+        positive_probabilities_product *= wordInfo.getSpamProbability();
+        negative_probabilities_product *= (1.0f - wordInfo.getSpamProbability());
+      }
+      double probabilities_products_sum = positive_probabilities_product + negative_probabilities_product;
+      return positive_probabilities_product / probabilities_products_sum;
+    }
+  }
+
+  static class LogProbabilityStrategy implements BayesianProbabilityCalculationStrategy {
```
oferb: Let's choose what we think is the best implementation - simple or log, and make is just a method in BayesFilter.

projects/noloan/app/filter/bayes/BayesFilter.java
```diff
@@ -0,0 +1,168 @@
+package projects.noloan.app.filter.bayes;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import projects.noloan.app.filter.SpamFilterAlgorithm;
+import projects.noloan.app.filter.bayes.Protos.*;
```
oferb: Let's avoid the * import.

projects/noloan/app/filter/bayes/BayesFilter.java
```diff
@@ -0,0 +1,168 @@
...
+   * calculate spam/ham rate and spam probability
+   */
+  private WordInfo wordInfoWithProbability(WordInfo wordInfo) {
+    float spamRate, hamRate, spamProbability = 0;
+
+    spamRate = wordInfo.getSpamCount() / (float) totalSpamCount;
+    hamRate = wordInfo.getHamCount() / (float) totalHamCount;
+    spamProbability = spamRate / (spamRate + hamRate);
+
+    if (spamProbability < 0.01f) {
```
oferb: Let's make these private static floats.

projects/noloan/app/filter/bayes/BayesFilter.java
```diff
@@ -0,0 +1,168 @@
...
+    return wordInfo
+            .toBuilder()
+            .setSpamRate(spamRate)
+            .setHamRate(hamRate)
+            .setSpamProbability(spamProbability)
+            .build();
+  }
+
+  private boolean calculateBayesianProbability(ArrayList<WordInfo> sms) {
+    return strategy.probability(sms) > 0.9f;
```
oferb: Let's make this a private static float.

projects/noloan/app/filter/bayes/BayesFilter.java
```diff
@@ -0,0 +1,168 @@
+package projects.noloan.app.filter.bayes;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import projects.noloan.app.filter.SpamFilterAlgorithm;
+import projects.noloan.app.filter.bayes.Protos.*;
+
+public class BayesFilter implements SpamFilterAlgorithm {
```
oferb: I don't think we need an interface. We have just one algorithm at this point.

projects/noloan/app/filter/bayes/BayesFilter.java
```diff
@@ -0,0 +1,168 @@
...
+      double nu = message
+              .stream()
+              .mapToDouble(word ->
+                      Math.log(1 - word.getSpamProbability()) -
+                              Math.log(word.getSpamProbability())).sum();
+      return 1/(1 + Math.exp(nu));
+    }
+  }
+
+  public boolean lineIsSpam(String line) {
```
oferb: How about we return the probability and let the caller decide?

oferb: s/lineIsSpam/isSpam

oferb: s/line/text
or
s/line/message

projects/noloan/app/filter/SpamFilter.java
```diff
@@ -4,20 +4,32 @@
 
 import com.google.common.io.Resources;
 import com.google.common.collect.ImmutableList;
+
```
oferb: How about if the spam filter adds the probability to the SmsMessage, for debugging?

projects/noloan/app/filter/bayes/BayesFilter.java
vmax: @oferb I view `BayesFilter` as a black box which gives binary answer — either it thinks a message is spam or it is not

projects/noloan/app/filter/SpamFilter.java
vmax: @oferb won't it break backwards compat with `.prototxt` we have?

vmax: @oferb we could `Log` it instead

projects/noloan/app/filter/SpamFilterAlgorithm.java
```diff
@@ -0,0 +1,5 @@
+package projects.noloan.app.filter;
```
oferb: I don't think we'll need 2 algorithms at the same time, we'll just evolve the one we have.
So no need for an interface.

PR #8, opened 2018-05-16
Make Storyteller read and write single proto

val-fed:


tools/storyteller/StoryWriter.java
```diff
@@ -18,70 +16,60 @@
...
-    String filename = timeString + "." + type;
-    try {
-      File statusFile = Paths.get(getUnsharedStoriesPath(), filename).toFile();
-      if (contents == null) {
-        statusFile.createNewFile();
-      } else {
-        fileUtils.writePrototxt(contents, statusFile.toString());
+    switch (status) {
+      case START: {
+        secondsShift = -1;
```
oferb: I think you should reset with `storyBuilder = Story.newBuilder();`, otherwise you get the leftovers from the previous story.

tools/storyteller/StoryWriter.java
```diff
@@ -18,70 +16,60 @@
...
+        storyBuilder.setProject(project).setStartTimeMs(getCurrentTimestamp(secondsShift));
+        if (timeMsLastSavedStoryItem == 0) {
+          timeMsLastSavedStoryItem = getCurrentTimestamp();
+        }
+        break;
+      }
+      case RUNNING: {
+        storyBuilder.setProject(project);
```
oferb: Please set `storyBuilder.setEndTimeMs(currentTime);` and save the story to file at every call to `writeStory`. This of this use case - 
You're running storyteller, working for an hour, and suddenly your computer shuts down. You want your current state to be saved.
Since the proto should be pretty small (tens of Kb), this shouldn't take much time.

tools/storyteller/StoryReader.java
```diff
@@ -29,110 +20,12 @@ public StoryReader(FileUtils fileUtils) {
...
-  private ImmutableList<FileData> getFileData(String path) throws ParseException {
-    ImmutableList.Builder<FileData> result = new ImmutableList.Builder<>();
-    if (!Paths.get(path).toFile().exists()) {
-      return result.build();
+  public ImmutableList<Story> getStories(String path) {
+    if (fileUtils.fileExists(path)) {
+      StoryList stories = (StoryList) fileUtils.readPrototxtUnchecked(path, StoryList.newBuilder());
+      return ImmutableList.copyOf(stories.getStoryList());
+    } else {
+      return ImmutableList.copyOf(StoryList.getDefaultInstance().getStoryList());
```
oferb: This is just an empty list, so `ImmutableList.of()`

tools/storyteller/StoryWriter.java
```diff
@@ -18,70 +16,64 @@
...
-        statusFile.createNewFile();
-      } else {
-        fileUtils.writePrototxt(contents, statusFile.toString());
+    switch (status) {
+      case START: {
+        timeMsLastSavedStoryItem = getCurrentTimestamp();
+        secondsShift = -1;
+        storyBuilder = Story.newBuilder();
+        storyBuilder.setProject(project).setStartTimeMs(getCurrentTimestamp(secondsShift));
+        if (timeMsLastSavedStoryItem == 0) {
```
oferb: You already set `timeMsLastSavedStoryItem = getCurrentTimestamp();`

tools/storyteller/StoryWriter.java
```diff
@@ -18,70 +16,64 @@
...
-    try {
-      File statusFile = Paths.get(getUnsharedStoriesPath(), filename).toFile();
-      if (contents == null) {
-        statusFile.createNewFile();
-      } else {
-        fileUtils.writePrototxt(contents, statusFile.toString());
+    switch (status) {
+      case START: {
+        timeMsLastSavedStoryItem = getCurrentTimestamp();
+        secondsShift = -1;
```
oferb: We actually don't need `secondsShift` anymore since we don't rely on filename timestamps.

tools/storyteller/StoryWriter.java
```diff
@@ -18,70 +16,64 @@
...
+        storyBuilder.setProject(project).setEndTimeMs(getCurrentTimestamp());
+        break;
+      }
+      case END: {
+        secondsShift = 1;
+        long currentTime = getCurrentTimestamp(secondsShift);
+        storyBuilder.setEndTimeMs(currentTime);
+        timeMsLastSavedStoryItem = currentTime;
+        storiesBuilder.addStory(storyBuilder.build());
+        fileUtils.writePrototxtUnchecked(
```
oferb: What if your computer shuts down before you reach here? Let's save in all cases.
This also means that you should close off any previous unclosed StoryItem with `setEndTimeMs()`

tools/storyteller/Storyteller.java
```diff
@@ -196,24 +203,19 @@ YearMonth getLastMonth(Instant instant) {
...
     writer.saveScreenshot(project, story);
   }
 
-  public ImmutableList<Story> getUnsharedStories() {
-    try {
-      return reader.getStories(getUnsharedStoriesPath());
-    } catch (ParseException e) {
-      throw new RuntimeException((e));
-    }
+  public StoryList getUnsharedStories() {
```
oferb: Let's change this also to `ImmutableList<Story>`

tools/storyteller/StoryWriter.java
```diff
@@ -18,70 +16,60 @@
...
-  private String getSharedStoriesPath() {
-    return Storyteller.getSharedStoriesPath(config);
-  }
-
-  private String getUnsharedStoriesPath() {
-    return Storyteller.getUnsharedStoriesPath(config);
-  }
-
-  public void writeStatusFile(FileData.Type type, Message contents) {
+  public void writeStory(Storyteller.StorytellerStatus status, String project) {
```
oferb: How about splitting this method into 3 parts, something like:
`startStory()`
`updateCurrentStory()`
`endStory()`

tools/storyteller/StoryWriter.java
```diff
@@ -18,70 +16,60 @@
...
-      } else {
-        fileUtils.writePrototxt(contents, statusFile.toString());
+    switch (status) {
+      case START: {
+        timeMsLastSavedStoryItem = getCurrentTimestamp();
+        storyBuilder = Story.newBuilder();
+        storyBuilder.setProject(project)
+                .setStartTimeMs(getCurrentTimestamp())
+                .setEndTimeMs(getCurrentTimestamp());
+        saveStories();
```
oferb: If `storyBuilder` is not in `storiesBuilder`, we're not saving it right?

tools/storyteller/StoryWriter.java
```diff
@@ -18,71 +17,70 @@
...
-  private String getCurrentTimeString() {
-    return getCurrentTimeString(0);
+  void saveStoryItem(String project, String story) {
+    // Replace empty with one space, so that the prototxt will have the entry
+    story = story.isEmpty() ? " " : story;
+    long currentTime = getCurrentTimestamp();
+    long timeMs = currentTime - lastSavedStoryItemTimeMs;
+    lastSavedStoryItemTimeMs = currentTime;
+    StoryItem storyItem = StoryItem.newBuilder().setTimeMs(timeMs).setOneliner(story).build();
+    ImmutableList<StoryItem> savedStoryItems = storyItems;
```
oferb: Isn't
```
ImmutableList<StoryItem> savedStoryItems = storyItems;
storyItems = ImmutableList.<StoryItem>builder().addAll(savedStoryItems).add(storyItem).build();
```

The same as this?
```
storyItems = ImmutableList.<>builder().addAll(storyItems).add(storyItem).build();
```

tools/storyteller/StoryWriter.java
```diff
@@ -18,71 +17,70 @@
...
+    storyItems = ImmutableList.of();
   }
 
-  private String getCurrentTimeString() {
-    return getCurrentTimeString(0);
+  void saveStoryItem(String project, String story) {
+    // Replace empty with one space, so that the prototxt will have the entry
+    story = story.isEmpty() ? " " : story;
+    long currentTime = getCurrentTimestamp();
+    long timeMs = currentTime - lastSavedStoryItemTimeMs;
```
oferb: Why are you setting `setTimeMs()` as `currentTime - lastSavedStoryItemTimeMs` and not just `currentTime`?

tools/storyteller/StoryWriter.java
```diff
@@ -18,71 +17,70 @@
...
 /* Storyteller writer that writes story files. */
 @Singleton
 public class StoryWriter {
   private static final Logger log = Logger.getForClass();
-  public static DateFormat DATE_FORMATTER = new SimpleDateFormat("yyyy-MM-dd_HH:mm:ss_z");
+  private static final DateFormat DATE_FORMATTER = new SimpleDateFormat("yyyy-MM-dd_HH:mm:ss_z");
 
   private Config config;
   private FileUtils fileUtils;
+  private StoryList.Builder storiesBuilder;
```
oferb: There seem to be too many local variables here..
I suspect you don't really need all of them.
Try just storing `StoryItem currentStoryItem`, `Story currentStory` and `List<Story> allStories (ArrayList)`, see if that works.

tools/storyteller/StoryWriter.java
val-fed: I misunderstood the value of this variable. I thought it was the duration of `item`.

tools/storyteller/StoryWriter.java
```diff
@@ -91,13 +84,36 @@ public void saveScreenshot(String project, String story) {
...
-      fileUtils.writePrototxt(
-          ScreenshotMetadata.newBuilder().setProject(project).setOneliner(story).build(),
-          filenameWithoutExtension + ".prototxt");
     } catch (AWTException | IOException e) {
       log.error("Error in saving screenshot", e);
     }
   }
+
+  private void updateCurrentStory(String project) {
+    currentStory = Story.newBuilder()
```
oferb: You can change currentStory to `Story.Builder` so you don't have to build every time.
Just so you know, you can turn any proto object into a builder with the current data.
So, `currentStory` is equal to `currentStory.toBuilder().build()`

tools/storyteller/StoryWriter.java
```diff
@@ -1,88 +1,82 @@
...
+    currentStoryBuilder
+        .setStartTimeMs(getCurrentTimestamp())
+        .setProject(project);
+    allStories.add(currentStoryBuilder.build());
+    saveStories();
   }
 
-  private String getUnsharedStoriesPath() {
-    return Storyteller.getUnsharedStoriesPath(config);
+  void updateStory(String project) {
```
oferb: The way the code currently is, you can call `updateStory()` without ever calling `startStory()`.
How about not setting `currentStoryBuilder` in the constructor (leaving it null), initializing it in `startStory()`, and if it's not initialized in `updateStory()` or `endStory()`, throwing an exception?

tools/storyteller/StoryWriter.java
```diff
@@ -91,13 +85,34 @@ public void saveScreenshot(String project, String story) {
...
-          ScreenshotMetadata.newBuilder().setProject(project).setOneliner(story).build(),
-          filenameWithoutExtension + ".prototxt");
     } catch (AWTException | IOException e) {
       log.error("Error in saving screenshot", e);
     }
   }
+
+  private void updateCurrentStory(String project) {
+    currentStoryBuilder
+        .setStartTimeMs(currentStoryBuilder.getStartTimeMs())
```
oferb: `.setStartTimeMs(currentStoryBuilder.getStartTimeMs())` is not doing anything.

tools/storyteller/StoryWriter.java
```diff
@@ -1,88 +1,88 @@
...
-        fileUtils.writePrototxt(contents, statusFile.toString());
-      }
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
-    if (type != FileData.Type.RUNNING) {
-      System.out.println(String.format("\n%s tracking: %s", type, timeString));
+  void updateStory(String project) {
+    if(currentStoryBuilder == null){
+      throw new RuntimeException("'currentStoryBuilder' is not initialized");
```
oferb: s/RuntimeException/IllegalStateException
IllegalStateException inherits from RuntimeException, but is a better fit.

PR #10, opened 2018-06-06
Fix bug with deleting unshared stories after reload storyteller

val-fed:


tools/storyteller/StoryWriter.java
```diff
@@ -41,11 +41,14 @@ public StoryWriter(StorytellerConfig storytellerConfig, FileUtils fileUtils) {
     allStories = new ArrayList<>();
   }
 
-  void startStory(String project) {
+  void startStory(String project, List<Story> stories) {
     currentStoryBuilder = Story.newBuilder();
     currentStoryBuilder
         .setStartTimeMs(getCurrentTimestamp())
         .setProject(project);
+    if(allStories.isEmpty()) {
```
oferb: Please add a space after `if`

tools/storyteller/StoryWriter.java
```diff
@@ -41,11 +41,14 @@ public StoryWriter(StorytellerConfig storytellerConfig, FileUtils fileUtils) {
     allStories = new ArrayList<>();
   }
 
-  void startStory(String project) {
+  void startStory(String project, List<Story> stories) {
```
oferb: Maybe add a separate method `addStories()` instead?

tools/storyteller/StoryWriter.java
```diff
@@ -50,16 +50,22 @@ void startStory(String project) {
     saveStories();
   }
 
+  void addStories(List<Story> stories){
+    if (allStories.isEmpty()) {
```
oferb: Why only if it's empty?

val-fed: `allStories` variable was not reset during the work of the Storyteller. And when we pressed the `pause` button `allStories` did not reset.
I changed behavior in the next commit. Now, regardless of whether Storyteller was closed completely or the pause button was pressed `allStories` is reset.

tools/storyteller/Storyteller.java
```diff
@@ -197,6 +197,7 @@ YearMonth getLastMonth(Instant instant) {
   }
 
   public void startup(String project) {
+    writer.addStories(getUnsharedStories());
```
oferb: Maybe you should put this in the constructor, so it's called only once. Would that fix the issue? Then you don't need to `allStories = new ArrayList<>();` in `addStories()`.

PR #11, opened 2018-06-14
Fix building problem and Add flogger

val-fed:


tools/local_server/LocalServer.java
```diff
@@ -21,27 +23,30 @@
...
 public class LocalServer {
-  private static final Logger logger = Logger.getLogger(LocalServer.class.getName());
+  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
   @FlagDesc(name = "local_server_port", description = "Port for local gRPC server")
   public static final Flag<Integer> localServerPort = Flag.create(8001);
 
   @FlagDesc(name = "root_path", description = "Root path for serving files for reviewer service")
-  public static final Flag<String> rootPath = Flag.create("");
+  public static final Flag<String> rootPath = Flag.create("/home/valerii/Work/storyteller/ok");
```
oferb: Should probably revert `/home/valerii/Work/storyteller/ok` to "".

tools/local_server/LocalServer.java
```diff
@@ -72,12 +77,17 @@ private static void checkFlags() {
     }
   }
 
+  @Singleton
+  @Component(modules = { CommonModule.class })
```
oferb: Good job!

tools/local_server/LocalServer.java
```diff
@@ -72,12 +77,17 @@ private static void checkFlags() {
...
+  public interface LocalServerComponent {
+    LocalServer getLocalServer();
+  }
+
   public static void main(String[] args) throws IOException, InterruptedException {
-    Flags.parse(args, LocalServer.class.getPackage(), CodeReviewService.class.getPackage(),
-        StorytellerService.class.getPackage());
+    Flags.parse(args, LocalServer.class.getPackage(), StorytellerService.class.getPackage());
```
oferb: I think you need to add the `AuthService`'s package (`google.startupos.tools.localserver.service`).

tools/local_server/LocalServer.java
val-fed: Sorry. I forgot to remove this.

tools/local_server/LocalServer.java
val-fed: Yes, it fixed the problem.

PR #13, opened 2018-07-06
Ci integration

vmax:


compile.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# Builds all targets and runs all tests
+if [[ ! -z "$CIRCLECI" ]]; then
+  DELETED_PACKAGES="";
+else
+  DELETED_PACKAGES="--deleted_packages $(cat .circleci/deleted_bazel_packages.txt)";
```
oferb: What is `deleted_bazel_packages`?

compile.sh
```diff
@@ -0,0 +1,13 @@
...
+fi
+
+PROTO_OPTIONS="--proto_compiler //external:proto_compiler --proto_toolchain_for_java //external:proto_java_toolchain"
+
+time bazel build $PROTO_OPTIONS $DELETED_PACKAGES //...
```
oferb: Why do we need `time`?

oferb: So now to normally compile (a single target), we need to add the `PROTO_OPTIONS` flag?

third_party/java_grpc_library.bzl
```diff
@@ -0,0 +1,133 @@
+def _path_ignoring_repository(f):
```
oferb: Where did this code come from?

compile.sh
vmax: @oferb oops, I think I blindly copy-pasted this from `startup-os` where it *is* needed for ignoring `cc_binary` target that would enforce full `protoc` compilation -- which we are trying to avoid

compile.sh
vmax: @oferb I needed it locally -- will remove

third_party/java_grpc_library.bzl
vmax: @oferb vendored-in [`java_grpc_library.bzl`](https://github.com/grpc/grpc-java/blob/master/java_grpc_library.bzl) with slight changes. Shall we provide a link?

compile.sh
vmax: @oferb yes. is it better to leave `@com_google_protobuf` in `WORKSPACE` or specify it in `README` (I prefer this one)?

third_party/java_grpc_library.bzl
```diff
@@ -1,3 +1,4 @@
+# SRC: https://github.com/grpc/grpc-java/blob/master/java_grpc_library.bzl
```
oferb: s/SRC/This code is adapted from:

WORKSPACE
```diff
@@ -98,8 +57,52 @@ maven_jar(
...
   artifact = "com.squareup.okio:okio:jar:1.6.0"
 )
-http_archive(
-    name = "com_google_protobuf",
-    urls = ["https://github.com/google/protobuf/archive/3.5.1.1.zip"],
-    strip_prefix = "protobuf-3.5.1.1",
+
+git_repository(
+    name = 'temp',
+    remote = 'https://github.com/vmax/temp',
```
oferb: Looks like you're still using your repo..

compile.sh
```diff
@@ -0,0 +1,16 @@
...
+RESET=$(tput sgr0)
+
+# Run with `bazel` command as param (build|test)
+if [[ $1 != "build" && $1 != "test" ]]; then
+  echo "$RED""Run script with 'build' or 'test' as param$RESET"
+  exit 1
+fi
+
+# Builds all targets and runs all tests
+PROTO_OPTIONS="--proto_compiler //external:proto_compiler --proto_toolchain_for_java //external:proto_java_toolchain"
```
oferb: Do we still need these here if they're in `bazelrc`?
Also, when should this script be run? Please document at the top.

vmax: @oferb I think we don't
You should run to regression-test all targets (build or test everything) -- will document

WORKSPACE
vmax: @oferb please downgrade plugin version in `startupos_binaries` so I could use those

fix-formatting.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
oferb: Should we apply the same fixes here? E.g have only fix-formatting, not run `buildifier` separately.

PR #17, opened 2018-08-07
Fix inability to save more than one file with shared stories

val-fed:


tools/storyteller/Storyteller.java
```diff
@@ -130,6 +132,11 @@ public void share() {
     }
   }
 
+  private String getSharedStoriesFilename() {
+    LocalDateTime currentTime = LocalDateTime.now();
+    return currentTime.format(DateTimeFormatter.ofPattern("yyyy-MM-dd_HH-mm-ss")) + ".prototxt";
```
oferb: Let's use the same filename as the screenshots.

PR #38, opened 2019-01-09
Mysadnaangular6

a3l5o7n9:
Styling changes to main components.


projects/mysadna2/src/app/app.component.html
```diff
@@ -1,10 +1,10 @@
+<nav class="app-navigation">
+  <a [routerLink]="'/projects'">Projects</a>
+  <a [routerLink]="'/users'">Users</a>
+</nav>
 <div style="text-align:center">
```
webdevelopland: Please, try always put css(scss) code in scss files only. Not in html.
The only exception is dynamic binding. Example:
```
<div [style.width.px]="blockWidth"></div>
```
But it's good to not use dynamic binding too if possible.

projects/mysadna2/src/app/app.component.scss
```diff
@@ -0,0 +1,16 @@
+.app-navigation {
+    float: right;
```
webdevelopland: 4 space intend style. Should be 2

projects/mysadna2/src/app/app.component.scss
```diff
@@ -0,0 +1,16 @@
...
+    >a {
+        margin: 5px;
+        width: 83px;
+        height: 24px;
+        font-family: OpenSans;
+        font-size: 18px;
+        font-weight: bold;
+        font-style: normal;
```
webdevelopland: You probably get all the `normal` styles from zeplin. But it's redundant. They do nothing and you can remove it.
I think the reason why zeplin show it, to reset all not normal values from another classes, which might be applies to the same element. But we don't have such issue.

projects/mysadna2/src/app/app.component.scss
```diff
@@ -0,0 +1,16 @@
+.app-navigation {
+    float: right;
+    >a {
+        margin: 5px;
+        width: 83px;
+        height: 24px;
```
webdevelopland: Again, you took the width and height from zeplin, but it's not the best idea in a real page.
Zeplin mock is just a static model, kind of a picture. But website is a more dynamic.
Try to open your site in Inspect mode (Ctrl+Shift+C)
The rules just can't be handled:
https://i.imgur.com/SDqZkT1.png

When you're creating a mockup, especially responsive design, try to make blocks flexible. When its size and style can be changed from inside and outside.

projects/mysadna2/src/app/app.component.scss
```diff
@@ -0,0 +1,16 @@
...
+        font-family: OpenSans;
+        font-size: 18px;
+        font-weight: bold;
+        font-style: normal;
+        font-stretch: normal;
+        line-height: normal;
+        letter-spacing: normal;
+        color: #000000;
+    }
+}
```
webdevelopland: end without new line

projects/mysadna2/src/app/user-details/user-details.component.html
```diff
@@ -1,34 +1,45 @@
...
+      <img src="{{user.userImg}}" class="userImage">
+      <div class="userName">{{user.userName}}</div>
+      <div class="email">{{user.email}}</div>
+      <ul class="social-Networks">
+        <li *ngFor="let socialNetwork of user.socialNetworks" class="socialNetwork">
+          <div class="{{socialNetwork}}">
+            <!-- svg -->
+            <img src="./assets/images/{{socialNetwork}}.svg" class="{{socialNetwork}}">
+            <!-- png -->
+            <!--<img src="./assets/images/linkedin.png" srcset="./assets/images/linkedin@2x.png 2x,./assets/images/linkedin@3x.png 3x" class="linkedin">-->
```
webdevelopland: Hardcoded code?

projects/mysadna2/src/app/user-details/user-details.component.html
```diff
@@ -1,34 +1,45 @@
...
         </li>
       </ul>
     </div>
-    <div class="skills">
-      <h4>Skills</h4>
-      {{user.skills}}
+    <div class="rightContainer">
+      <div class="projects">
+        <h4>Projects:</h4>
+        <ul class="projects-list">
```
webdevelopland: :+1:

projects/mysadna2/src/app/user-details/user-details.component.scss
```diff
@@ -0,0 +1,69 @@
+.userHeader {
```
webdevelopland: Usually class names (and all css variable) are used in kebab-case:
```
.user-header {
````

File: projects/mysadna2/src/app/app.component.scss (hunk 3 above)
Comment (a3l5o7n9) [reply projects/mysadna2/src/app/app.component.scss]: Do you mean I can simply remove all 'px' instances?

File: projects/mysadna2/src/app/app.component.scss (hunk 5 above)
Comment (a3l5o7n9) [reply projects/mysadna2/src/app/app.component.scss]: Already fixed in more recent commit yet to have been uploaded.

File: projects/mysadna2/src/app/user-details/user-details.component.html (hunk 6 above)
Comment (a3l5o7n9) [reply projects/mysadna2/src/app/user-details/user-details.component.html]: Yes.

Comment (a3l5o7n9) [reply projects/mysadna2/src/app/user-details/user-details.component.html]: I didn't know what the png option meant, so I left it there in a comment for the moment.

File: projects/mysadna2/src/app/app.component.scss (hunk 3 above)
Comment (webdevelopland) [reply projects/mysadna2/src/app/app.component.scss]: No. I mean you can remove this:
```css
font-style: normal;
font-stretch: normal;
line-height: normal;
letter-spacing: normal;
```

a3l5o7n9: Implemented and pushed commit.

projects/mysadna2/src/app/app.component.scss
```diff
@@ -1,16 +1,20 @@
...
-        line-height: normal;
-        letter-spacing: normal;
-        color: #000000;
-    }
-}
\ No newline at end of file
+  float: right;
+  >a {
+    margin: 5px;
+    width: 83;
```
webdevelopland: I guess you got me wrong.
The string `width: 83` is invalid. You have to set unit (px, em, pt etc).

In my review I meant: if size of block can be changed from inside, then you shouldn't set size of the block.
Size of `a` element depends on text inside. width and height is redundant here.

a3l5o7n9: So what can I do to keep the proportions offered in the mockups?

PR #43, opened 2019-01-14
NoLoan DataUploader 

gulilo:


projects/noloan2/server/DataUploader.java
```diff
@@ -0,0 +1,8 @@
+package hasadna.projects.noloan2.server;
+
+public class DataUploader {
+
```
oferb: Please reintroduce the code here, except the part inside `run()`:
https://github.com/google/startup-os/blob/master/common/Base64Tool.java#L42-L71

We'll be using it later.

projects/noloan2/server/DataUploader.java
```diff
@@ -0,0 +1,38 @@
...
+    private FileUtils fileUtils;
+
+    @Inject
+    DataUploader(FileUtils fileUtils) {
+        this.fileUtils = fileUtils;
+    }
+
+    void run() throws IOException {
+      //TODO implament
```
oferb: `s/implament/implement`

PR #69, opened 2019-01-21
Startup script

oferb:


tools/startup/setup.sh
```diff
@@ -0,0 +1,114 @@
...
+sudo apt-get purge -y ubuntu-web-launchers
+
+# Copy wifi credentials
+mkdir -p wifis
+sudo cp /etc/NetworkManager/system-connections/* wifis/
+sudo cp wifis/* /etc/NetworkManager/system-connections/
+sudo systemctl restart NetworkManager
+
+# Download apt installs if needed
+# TODO: Figure out how to download packages from a predefined list, instead of "latest". This list will be generated from "latest" at some point in time.
```
vmax: @oferb is [this](https://www.pcsuggest.com/apt-generate-package-download-list/) something that you had in mind?

tools/startup/setup.sh
```diff
@@ -0,0 +1,115 @@
...
+# Set up StartupOS:
+git clone https://github.com/google/startup-os.git downloads/startup-os
+cd downloads/startup-os
+bazel run //tools/reviewer/aa:aa_tool -- init --base_path /media/$USER/data/base
+cd /media/$USER/data/base/head/startup-os
+./test.sh
+# FIXME: Why does it not add to PATH:
+source ~/.bashrc
+aa
+echo "Please sign-in at https://web-login-startupos.firebaseapp.com and then check that ~/.aa_token exists"
```
vmax: @oferb from this line I understand your goal is to start `localserver`. What if we did it explicitly (put smth like [these lines](https://github.com/google/startup-os/blob/master/tools/reviewer/aa/aa_tool.sh#L126-L127))?

tools/startup/setup.sh
oferb: Something along those lines. I'd like to get a reproducible setup.

tools/startup/setup.sh
oferb: We could, but why?
To avoid the "Unknown command"?

tools/startup/setup.sh
```diff
@@ -0,0 +1,115 @@
...
+  mkdir downloads/apt
+  curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > downloads/microsoft.gpg
+  sudo apt-get -dy update
+  sudo apt-get -fdy install git default-jdk pkg-config zip g++ zlib1g-dev unzip python python3 curl apt-transport-https
+  sudo apt-get -dy update
+  wget -O downloads/gconf2-common_3.2.6-4ubuntu1_all.deb http://archive.ubuntu.com/ubuntu/pool/universe/g/gconf/gconf2-common_3.2.6-4ubuntu1_all.deb
+  wget -O downloads/libgconf-2-4_3.2.6-4ubuntu1_amd64.deb http://archive.ubuntu.com/ubuntu/pool/universe/g/gconf/libgconf-2-4_3.2.6-4ubuntu1_amd64.deb
+  sudo apt-get install -fdy code
+  sudo apt-get -dy upgrade
+  # TODO: Figure out how to cache the "apt-get update" to the downloads folder.
```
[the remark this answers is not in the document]
oferb: Yup, good link :)

tools/startup/setup.sh
vmax: @oferb yes. also, explicit is better than implicit

oferb: Cool, added a TODO.

PR #77, opened 2019-01-24
code review, there is an error 

yehudalev:
error: exception ParseException is never thrown in body of corresponding try statement
    } catch (IOException | ParseException e) {

### DIFF
--- a/projects/opentrain/gtfs_pipeline/data_handler/DataHandler.java
+++ b/projects/opentrain/gtfs_pipeline/data_handler/DataHandler.java
@@ -1,4 +1,23 @@
 package projects.opentrain.gtfs_pipeline.data_handler;
+
+//imports for proto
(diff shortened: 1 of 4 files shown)


projects/opentrain/gtfs_pipeline/data_handler/DataHandler.java
```diff
@@ -1,4 +1,23 @@
...
+
+//imports for proto
+import com.projects.opentrain.gtfs_pipeline.StopsProtos.Stop;
+import com.projects.opentrain.gtfs_pipeline.StopsProtos.Stops;
+import com.google.protobuf.TextFormat;
+import java.io.Reader;
+
+//imports for Appache
+import org.apache.commons.csv.CSVFormat;
+import org. apache.commons.csv.CSVParser;
```
val-fed: Looks like there is reductant whitespace.

PR #94, opened 2019-02-01
process the stackoverflow Posts.xml file: initial code.

### LINKED ISSUE #68: Process the 12GB Stackoverflow data file and strip out unneeded data
This task is just to reduce the size of the huge 12GB file, by removing the fields we don't need. I expect the text fields take most space (the question, answers etc).
The challenge here is to be able to work with a 12GB file. You will probably not be able to load it all into memory, so you will need to process it in a streaming manner.

Here's some material from the internet on dealing with large files:
https://www.baeldung.com/java-read-lines-large-file
https://stackoverflow.com/questions/8813982/reading-and-processing-big-text-file-of-25gb
https://stackoverflow.com/questions/26310595/how-to-parse-big-50-gb-xml-files-in-java

Another option is to somehow break it up into several smaller files, and process these in memory. Not sure if that's easily possible with these XML files as they may have headers and footers.

mohammad-suliman:
### link to issue:
fixes #68 
### known issues with pull request:
* the paths of the input and output files that the code reads from and writes the extracted data to are still hard coded in the main function of DataExtractor class. additional work is needed to make getting those paths similar to the way other projects in the repo get such input.
* there is no build file in the project which configures it to be built with compile.sh script. such build configuration file is needed.
note: the project compiles fine with javac command.
* regarding extracted file size: the file has been reduced from nearly 63 GB to 5 GB. however this is not sufficient according to the conversation with @oferb . so additional reduction is needed, this will be in a future PR.
* last thing I wanted to mention for the sake of completeness, the original file which the code reads the data from (Posts.xml) has a row tag of the format:
<row ...attributes.../>
the extracted file (Posts-extracted.xml) in our case has a row tag of the format:
<row ...extracted attributes...></row>
semantically, both formats are the same, but syntactically they are different. this difference was as a result of a limitation of the API used in this PR.
I mentioned this change because both files are actually a data base table with a huge number of row tags.


projects/stackoverflow/src/stackoverflow/DataExtractor.java
```diff
@@ -0,0 +1,178 @@
+package stackoverflow;
```
oferb: Please place this file directly under `projects/stackoverflow`, and change the package to `projects.stackoverflow`.

projects/stackoverflow/src/stackoverflow/DataExtractor.java
```diff
@@ -0,0 +1,178 @@
...
+import javax.xml.stream.events.XMLEvent;
+
+public class DataExtractor {
+
+	// paths of the files that we read from and write to
+	private String inputFilePath;
+	private String outputFilePath;
+
+	// the name of the row tag.
+	private static final String ROW = "row";
```
oferb: I'd just rename this to `ROW_TAG` and then remove the comment.
Similarly for `POSTS`.

projects/stackoverflow/src/stackoverflow/DataExtractor.java
```diff
@@ -0,0 +1,178 @@
...
+	// the name of the posts tag
+	private static final String POSTS = "posts";
+	// some constant strings for the relevant attributes's names in the row tag
+	private static final String ID = "Id";
+	private static final String Post_TYPE_ID = "PostTypeId";
```
oferb: Capital letters: `s/Post_TYPE_ID/POST_TYPE_ID`

projects/stackoverflow/src/stackoverflow/DataExtractor.java
```diff
@@ -0,0 +1,178 @@
...
+	// paths of the files that we read from and write to
+	private String inputFilePath;
+	private String outputFilePath;
+
+	// the name of the row tag.
+	private static final String ROW = "row";
+	// the name of the posts tag
+	private static final String POSTS = "posts";
+	// some constant strings for the relevant attributes's names in the row tag
+	private static final String ID = "Id";
```
oferb: Maybe add `_ATTRIBUTE` to the names of the constants, and then simplify the comment to:
`// ROW_TAG attributes:`

projects/stackoverflow/src/stackoverflow/DataExtractor.java
```diff
@@ -0,0 +1,178 @@
...
+	private static final String Post_TYPE_ID = "PostTypeId";
+	private static final String CREATION_DATE = "CreationDate";
+	private static final String OWNER_USER_ID = "OwnerUserId";
+	private static final String TAGS = "Tags";
+
+	public void extractData() {
+		// deffine those variables here so they can be cleaned in finally block
+		XMLEventReader eventReader = null;
```
oferb: You can use try-with-resources to avoid having to initialize and then close:
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

Note that you can have a single try with both resources.

projects/stackoverflow/src/stackoverflow/DataExtractor.java
```diff
@@ -0,0 +1,178 @@
...
+
+	// the name of the row tag.
+	private static final String ROW = "row";
+	// the name of the posts tag
+	private static final String POSTS = "posts";
+	// some constant strings for the relevant attributes's names in the row tag
+	private static final String ID = "Id";
+	private static final String Post_TYPE_ID = "PostTypeId";
+	private static final String CREATION_DATE = "CreationDate";
+	private static final String OWNER_USER_ID = "OwnerUserId";
```
oferb: General comment - looks like you're using tabs. Can you equally have used 2 spaces instead, or is it more difficult in some way?

projects/stackoverflow/src/stackoverflow/DataExtractor.java
mohammad-suliman: Unfortunately, it seems that try-with-resources won't help us here. The classes XMLEventReader and XMLEventWriter doesn't seem to implement java.lang.AutoCloseable interface, so the try-with-resources can auto close them.

oferb: Ok, got it.

projects/stackoverflow/ExtractedRow.java
```diff
@@ -0,0 +1,57 @@
+package projects.stackoverflow;
+
+public class ExtractedRow {
```
oferb: I suggest that instead of this class, you use a proto (Protocol Buffers).
You can see an example of a proto here:
https://github.com/hasadna/hasadna/blob/master/projects/opentrain/gtfs_pipeline/common/common.proto

It has getters and setters, and also a `writeDelimitedTo` that writes directly to a file stream:
https://github.com/hasadna/hasadna/blob/master/projects/opentrain/gtfs_pipeline/data_handler/DataHandler.java#L56

So using it, you also won't need `writeExtractedRowToOutput()` and `writeEndOfLine()`.

You can also later read them from the file directly to an object.

You'll need to define 2 build targets:
https://github.com/hasadna/hasadna/blob/master/projects/opentrain/gtfs_pipeline/common/BUILD

Then, you can depend on them from your java file to get the proto imports:
https://github.com/hasadna/hasadna/blob/master/projects/opentrain/gtfs_pipeline/data_handler/BUILD#L13

If you prefer, I can approve this PR and you do it in a subsequent PR.

PR #95, opened 2019-02-02
a tool for downloading APT dependencies

vmax:


tools/startup/apt-deps
```diff
@@ -0,0 +1,38 @@
...
+      exit 1;
+    fi
+
+    echo "Downloading to debs/"
+    mkdir -p debs/
+    wget --directory-prefix=debs/ --input-file=${infile}
+elif [[ "$verb" = "install" ]]; then
+    echo "Installing debs from directory debs/"
+
+    dpkg --force-depends -i debs/*.deb
```
oferb: Can't find --force-depends in the man: https://linux.die.net/man/1/dpkg

tools/startup/apt-deps
```diff
@@ -0,0 +1,38 @@
...
+    outfile=$3
+
+    if [[ ! -f "$infile" ]] || [[ -z "$outfile" ]]; then
+      echo "Should supply input and output files: $0 $1 [input] [output]"
+      exit 1;
+    fi
+
+    echo "Running apt-get update, installing wget"
+    apt-get update &> /dev/null
+    apt-get install -y wget &> /dev/null
```
oferb: Do you think we could easily make it so that if `wget` is cached on the disk it takes it from there?

tools/startup/apt-deps
```diff
@@ -0,0 +1,38 @@
...
+    echo "Generating ${outfile}"
+    apt-get -qq -y install --reinstall --print-uris $(cat ${infile}) | awk $'{ gsub(/\'/, "", $1); print $1}' >  ${outfile}
```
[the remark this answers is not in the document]
vmax: @oferb `--print-uris` already implies not downloading files, but only printing a list of URLs to download; it doesn't actually do any installations.

tools/startup/apt-deps
vmax: @oferb I don't think it's needed. On the majority of distributions `apt-get install -y wget` will be a `no-op`. I only included it because I tested the script on Docker container, which is stripped from almost _everything_  🙁

tools/startup/apt-deps
```diff
@@ -0,0 +1,38 @@
...
+elif [[ "$verb" = "download" ]]; then
+    echo "Downloading debs from file ${infile}"
+
+    infile=$2
```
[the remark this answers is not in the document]
vmax: @oferb you mean, storing `$2` as `$infile`? strictly speaking, we don't need to do it. i think, it brings in some semantic meaning, but we could remove it

tools/startup/apt-deps
vmax: @oferb I found it on the internets. what it essentially does is just installs all packages you give it, not minding `Depends:` field. I had some troubles with installing packages without it, but we could remove it & experiment.

PR #96, opened 2019-02-09
Setup script should exit if no connection is available

vmax:


tools/startup/setup.sh
```diff
@@ -32,8 +32,19 @@ sudo apt-get purge -y ubuntu-web-launchers
...
 sudo cp /etc/NetworkManager/system-connections/* wifis/
 sudo cp wifis/* /etc/NetworkManager/system-connections/
+echo "Restarting system network service"
 sudo systemctl restart NetworkManager
 
+echo "Waiting for internet connection to settle"
+sleep 10
+
+echo "Checking for internet connection"
+http_code=$(curl -LI httpbin.org/get -o /dev/null -w '%{http_code}\n' -s)
```
oferb: Why `httpbin.org`? :-)

vmax: @oferb first thing that popped into my mind 🙂

PR #106, opened 2019-02-15
Mysadna - responsive update

webdevelopland:
What's new:
- responsive design update
- new tslint rules


projects/mysadna/src/app/app.component.scss
```diff
@@ -1,56 +1,40 @@
...
+$logo-size: 74px;
+.app-logo {
+  position: absolute;
+  top: 5px;
+  left: 5px;
+  justify-self: flex-start;
+  display: flex;
+  >.my-sadna-logo {
+    height: $logo-size;
+    width: $logo-size;
```
webdevelopland: Here's an example of using scss variables.

projects/mysadna/src/app/app.component.scss
```diff
@@ -1,56 +1,40 @@
...
-    @media(max-width: 575.98px) {
-      align-self: center;  
-    }
+    padding: 18px;
     >.app-navigation {
       >a {
-        margin: 1em;
-        font-size: 1.25em;
+        margin: 20px;
+        font-size: 18px;
```
webdevelopland: I prefer to avoid relative font sizes, especially if we have mock in px

projects/mysadna/src/app/components/project-details/project-details.component.html
```diff
@@ -1,29 +1,18 @@
-<div *ngIf="project" class="project-details">
-  <h1 class="project-name">{{project.getProjectName()}}</h1>
-  <h3 class="project-description">{{project.getProjectDescription()}}</h3>
-  <a href="{{project.getProjectSite()}}" target="blank" class="project-link">
-    <h3>
+<div class="project-details media" *ngIf="project">
+  <div class="project-name">{{project.getProjectName()}}</div>
```
webdevelopland: I prefer to not use h tag, because the tag is basic `div`, but with a lot of excess styles.

projects/mysadna/src/app/components/project-details/project-details.component.ts
```diff
@@ -12,36 +12,28 @@ import { FirebaseService } from '@/services';
...
+    this.loadData(projectId);
   }
 
-  goBack(): void {
-    this.location.back();
+  loadData(projectId: number): void {
+    zip(
+      this.firebaseService.getUserList(),
+      this.firebaseService.getProjectList()
+    ).subscribe(data => {
```
webdevelopland: Here's we can use the sync way, which we discussed.
To not create queue, we can load users and projects at the same time, and subscribe on changes in one place.

projects/mysadna/src/app/components/user-details/user-details.component.html
```diff
@@ -1,51 +1,47 @@
...
-          <li *ngFor="let project of userProjects">
-            <a [routerLink]="'/project-details/' + project.getProjectId()" class="project">
-              {{project.getProjectName()}}
+<div class="user-details" *ngIf="user">
+  <div class="media">
+    <div class="user-header">{{ user.getFirstName() }} {{ user.getLastName() }}</div>
+    <div class="user-container media-row-col">
+      <div class="user-label">
+        <div class="flex">
+          <img [src]="user.getUserImg()">
```
webdevelopland: Instead of using expression binding (`{{ expression }}`), we can bind whole attribute: `[src]`.

projects/mysadna/src/app/services/firebase.service.ts
```diff
@@ -31,72 +26,26 @@ export class FirebaseService {
...
-    // Convert binary to project
-    const project: Project = Project.deserializeBinary(binary);
-
-    return project;
+  // Converts firebaseElement to binary
+  private getBinary(firebaseElement: FirebaseElement): Uint8Array {
+    const binary: Uint8Array = this.encodingService.decodeBase64StringToUint8Array(
+      firebaseElement.proto
+    );
+    return binary;
```
webdevelopland: Here's what I meant by "to make it more DRY"

projects/mysadna/src/app/shared/contributor/contributor.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component, Input } from '@angular/core';
+
+import { User } from '@/proto';
+
+@Component({
+  selector: 'app-contributor',
```
webdevelopland: Instead of create same code several times, we can create shared component with the code

projects/mysadna/src/styles/media.scss
```diff
@@ -0,0 +1,69 @@
...
+  .media {
+    width: 100%;
+  }
+  .media-row-col {
+    flex-direction: row;
+  }
+  .media-small-block {
+    display: none !important;
+  }
+}
```
webdevelopland: Here's media settings. I understand, that's pretty confusing. But let me explain what the code does.
It creates classes, which behave different, depends on screen size.
E.g. `media-row-col` put blocks into a column, if screen is small.

projects/mysadna/tslint.json
```diff
@@ -1,131 +1,46 @@
...
-    "codelyzer"
-  ],
   "rules": {
-    "arrow-return-shorthand": true,
-    "callable-types": true,
-    "class-name": true,
+    "eofline": true,
+    "curly": true,
+    "no-unused-variable": true,
+    "max-line-length": [true, 100],
```
webdevelopland: New tslint rules :blush:

projects/mysadna/src/app/app.component.scss
a3l5o7n9: I understand how to use them a little better now, thanks.

projects/mysadna/src/app/app.component.scss
a3l5o7n9: Understood.
How do you know that the sizes in the mock won't differ between two different screens?

projects/mysadna/src/app/components/project-details/project-details.component.html
a3l5o7n9: Noted.

projects/mysadna/src/app/components/project-details/project-details.component.ts
a3l5o7n9: Does `zip` create an array the length of which equals the number of calls made within the `zip` and then insert the return value of each call into the corresponding index?

projects/mysadna/src/app/components/user-details/user-details.component.html
a3l5o7n9: Noted.

projects/mysadna/src/app/services/firebase.service.ts
a3l5o7n9: I didn't even think of that way.

projects/mysadna/src/app/components/project-details/project-details.component.ts
webdevelopland: @a3l5o7n9 Exactly!

projects/mysadna/src/app/app.component.scss
webdevelopland: @a3l5o7n9 It will differ and it's good.
I mean if font is 15px then it's going to be 15px on all screens, but it may look different, depends on screen resolution. Sometimes it's ok and it's the planed. Sometimes not, and if need we can font size can be dependent on device screen. But still we don't need to use em anyway, em is just more easiest way in some cases.

PR #110, opened 2019-02-18
Shared protoc

webdevelopland:
- shared protoc, which can be used for any project
- new protoc can work with proto from startup-os


projects/angular-proto-firestore/README.md
```diff
@@ -31,10 +31,7 @@ How to use firebase functions:
 https://firebase.google.com/docs/functions/get-started  
 
 ## Updating protos
-After every proto update, run `npm run protoc`.
-
-## Supported platforms
-Development is supported on Linux and Mac only (`npm run protoc` doesn't support Windows).
+You need to create new proto functions after each update of .proto file.
```
oferb: `s/proto functions/proto functions (see above)`

common/PROTOC.md
```diff
@@ -0,0 +1,11 @@
+# How to create proto functions
+
+## Installation
+Run `npm install` to install node modules
+
+## Running
+Run `npm run protoc <relative path to your project>` to generate proto functions.
```
oferb: 1. So no gulp anymore? What did we gain, what did we lose?
2. Can we use this to create protos from the `startup-os` repo, assuming it's in the parent folder?
3. Do you think we could merge this with this? https://github.com/google/startup-os/blob/master/tools/reviewer/webapp/gulpfile.js

webdevelopland: @oferb 
1. gulp is nodejs script too, but sometimes it's more convenient than pure nodejs, because it has own extra methods. e.g. gulp clean, which is able to remove all files in a folder. And much more..
I decided to not use gulp, because it doesn't work well with attributes. Which mean we aren't able to use command like that: `npm run protoc projects/angular-proto-firestore`, because `projects/angular-proto-firestore` is interpret as name of gulp script.

webdevelopland: 2. I'm not sure what you mean. The script imports .proto files from parent `startup-os` and creates proto functions inside a direct project
https://github.com/hasadna/hasadna/blob/WD9/common/protoc.js#L64
Also yeah.. you can run that command:
```
npm run protoc projects/startup-os
```
And it's going to work, if `startup-os` is contained in a repo. But I guess there's no sense in the action.

webdevelopland: 3. Sure. But we need to update this script at first. protoc script from reviewer can work with imports, but the one can't.

oferb: 1. Great :) that's what I thought, but wanted to get your answer.
2a. How does it do that? Relative path through parent folder?
2b. How could startup-os be contained in hasadna repo?
2c. Not sure I follow that one, but I think that's what I meant.
The use-case is this:
> We're working on the profile pages, which are in hasadna, and want to use the startup-os protos. Could we do that?

3. Not sure I follow. You mean proto imports? If the one in startup-os can, then maybe we can just use it?

PR #115, opened 2019-02-19
Create LawsuitPdf class

cotom972:
1. Reads a lawsuit template in an Html format.
2. Updates lawsuit form with spam company's & user's details.
3. Uses iText7 to convert the resulted form to a PDF file.
4. Saves the file on device's storage.


projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
```diff
@@ -0,0 +1,248 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.TimeZone;
+
+import noloan.R;
+
+public class LawsuitPdf extends AppCompatActivity { // TODO: Add logs to class functions.
```
oferb: Let's rename to `LawsuitPdfActivity`
From now on, renaming will be written as:
`s/LawsuitPdf/LawsuitPdfActivity`
`s` is for switch.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
```diff
@@ -0,0 +1,248 @@
...
+
+
+    private static final String TAG = "Lawsuit";
```
oferb: How about 
`s/Lawsuit/LawsuitPdfActivity`? Is it too long?

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
```diff
@@ -0,0 +1,248 @@
...
+
+    // Folder configurations:
+    // TODO: Check if preferable to move these to an external folder configuration file.
+    private static String appName = "";                                             // Get App's name as main dir.
+    private static final String lawsuitPath = "/Lawsuits/";                         // General lawsuits dir.
```
oferb: static finals should be UPPER_SNAKE_CASE.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
```diff
@@ -0,0 +1,248 @@
...
+    private static final String DATE_FORMAT = "dd-M-yyyy";
+    private static final String TIME_ZONE = "Asia/Jerusalem";
+    private static SimpleDateFormat dateAndTimeFormatter;
+    private static SimpleDateFormat dateFormatter;
+
+    // Template's inner text (been read on create from template's file):
+    private static String templateContent = "";
+
+// ------------------------------- Lawsuit Form Fields ------------------------------- //
+//  TODO: Check if preferable to move this to an external file holding constant String values of lawsuit form.
```
oferb: No, it's fine to leave it here.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
```diff
@@ -0,0 +1,248 @@
...
+
+import noloan.R;
+
+public class LawsuitPdf extends AppCompatActivity { // TODO: Add logs to class functions.
+
+
+    private static final String TAG = "Lawsuit";
+
+    // Folder configurations:
+    // TODO: Check if preferable to move these to an external folder configuration file.
```
oferb: I think it's fine to leave it here.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
```diff
@@ -0,0 +1,248 @@
...
+    // Template's inner text (been read on create from template's file):
+    private static String templateContent = "";
+
+// ------------------------------- Lawsuit Form Fields ------------------------------- //
+//  TODO: Check if preferable to move this to an external file holding constant String values of lawsuit form.
+    private String spamType = "הודעה אלקטרונית";        // Type: SMS/Phone call/Email.
+    private String claimAmount = "סכום תביעה";
+    private Date receivedSpamDate;                    // Date received the spam.
+    private Date currentDate;                         // Current date of lawsuit creation (used for lawsuit signature).
+    private boolean sentHaser = true;                 // Indicates whether user has sent "haser" or not.
```
oferb: I'd shorten by removing the ` or not`.
Whether + the boolean field are clear enough IMO.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
```diff
@@ -0,0 +1,248 @@
...
+    private boolean sentMoreThanFiveLawsuits = false; // Indicates whether user has sent more then 5 lawsuits already in the past year or not.
+    private String moreThanFiveLawsuits = sentMoreThanFiveLawsuits ? "הגיש" : "לא הגיש/ו";
+    private String claimCase = sentHaser ? "לאחר שהתובע חזר בו מהסכמתו לקבלת דבר/י פרסומת מהנתבע/ים, על-ידי משלוח הודעת סירוב כהגדרתה בחוק" : "למרות שח\"מ לא נתן את הסכמתו המפורשת מראש לקבלת דבר/י הפרסומת";
```
oferb: `claimCase` is evaluated when the class is instantiated, and so is `sentHaser`, which is set to `true`, so wouldn't `claimCase` always get the same value?

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
```diff
@@ -0,0 +1,248 @@
...
+    private String companyFax = "פקס חברה";
+
+    // Second spam company's details:
+    private String company2Name = "שם החברה 2";
+    private String company2Id = "מספר ח.פ 2";
+    private String company2Address = "כתובת החברה 2";
+    private String company2Phone = "טלפון החברה 2";
+    private String company2Fax = "פקס חברה 2";
+// -------------------------------------------------------------------------------------:
+    Button btnCreate;
```
oferb: We prefer to use full names instead of shortened ones. So I suggest, `s/btnCreate/createButton`.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
```diff
@@ -0,0 +1,248 @@
...
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_lawsuit_pdf);
+        btnCreate = (Button) findViewById(R.id.button_createPDF);
```
oferb: s/button_createPDF/button_create_pdf
To avoid snake_case and CamelCase mix.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
```diff
@@ -0,0 +1,248 @@
...
+        appName = getString(R.string.app_name); // Get App's name
```
oferb: I think `// Get App's name` doesn't add meaningful information to the code, so I suggest to remove it.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
```diff
@@ -0,0 +1,248 @@
...
+
+        // Check Read/Write permissions:
+        isWriteStoragePermissionGranted();
+
+        // Read lawsuit template's content:
+        templatePath = Environment.getExternalStorageDirectory().getPath() + "/" + appName + "/" + templatePath;
```
oferb: Please try to replace all `"/"` with a path joining method like:
`Paths.get(a, b, c).toString()`
https://developer.android.com/reference/java/nio/file/Paths

This avoids potential errors with missing or redundant `"/"`'s.

projects/noloan2/app/src/main/res/raw/itextkey.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<license date="2019-02-05" issuer="iText Group NV" library="3">
+    <licensee expire="2019-03-05">
```
oferb: Expire on `2019-03-05`?

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
```diff
@@ -0,0 +1,248 @@
...
+
+        // Update date values:
+        dateAndTimeFormatter = new SimpleDateFormat(DATE_TIME_FORMAT);
+        dateAndTimeFormatter.setTimeZone(TimeZone.getTimeZone(TIME_ZONE));
+        dateFormatter = new SimpleDateFormat(DATE_FORMAT);
+        dateFormatter.setTimeZone(TimeZone.getTimeZone(TIME_ZONE));
+
+        // Create lawsuit's file name with current date:
+        String directory_path = Environment.getExternalStorageDirectory().getPath() + lawsuitOutput;
+        Date dt = new Date();
```
oferb: You can just use `new Date()` inline:
`String targetPdf = directory_path + dateAndTimeFormatter.format(new Date()) + ".pdf";`

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
cotom972: Done.

projects/noloan2/app/src/main/res/raw/itextkey.xml
cotom972: Yes. This is the trial iText give for the "typography" module which fixes the Hebrew. 
https://itextpdf.com/en/resources/faq/legal-questions/itext-5/itext-java-library-free-charge-or-are-there-any-fees-be-paid

From what I understood - if used as open source & free app, legit to keep on using it. 
I can't say I'm a 100% sure for this. Still need to check, if you think it's worth. it.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,327 @@
...
+        // app_name/lawsuit_output_dir/
+        lawsuitOutputPath = Paths.get(Environment.getExternalStorageDirectory().getPath(),getString(R.string.app_name)+" - "+ LAWSUIT_OUTPUT_DIR_NAME).toString();
+        if (!dirExists(lawsuitOutputPath)){
+            File directory = new File(lawsuitOutputPath);
+            directory.mkdirs();
+        }
+    }
+
+    /**
+     * Cehcks if a directory given in the parameter exists.
```
oferb: s/Cehcks/Checks

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,327 @@
...
+
+        return res;
+    }
+
+    /**
+     * Genereates the lawsuit Pdf file name according to Time & Date user had confirmed the form.
+     */
+    public void createPdfLawsuitFileName(){
+
+        lawsuitOutputPdfFileName = dateAndTimeFormatter.format(new Date()) + ".pdf";
```
oferb: Looks like `lawsuitOutputPdfFileName` is not used outside this method, so you can make it a local, or even better I think, not use it at all:
`lawsuitOutputPdfFilePath = Paths.get(lawsuitOutputPath, dateAndTimeFormatter.format(new Date()) + ".pdf").toString();`

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,327 @@
...
+        dateFormatter = new SimpleDateFormat(DATE_FORMAT);
+        dateFormatter.setTimeZone(TimeZone.getTimeZone(TIME_ZONE));
+
+        // Generate lawsuit PDF OnClick:
+        button_create_pdf.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View view) {
+                try {
+                    createPdfLawsuitFileName();
+                    createPdf();
```
oferb: It seems to me that, after your changes to this PR, `createPdfLawsuitFileName()` will just set `lawsuitOutputPdfFilePath`, so you can simplify the code by not having such a member, and just calling the method straight from `createPdf()` into a local method variable inside `createPdf()`.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,327 @@
...
+    private String companyFax = "פקס חברה";
+
+    // Second spam company's det ails:
+    private String company2Name = "שם החברה 2";
+    private String company2Id = "מספר ח.פ 2";
+    private String company2Address = "כתובת החברה 2";
+    private String company2Phone = "טלפון החברה 2";
+    private String company2Fax = "פקס חברה 2";
+// ----------------------------- - -------------------------------------------------------:
+    Button button_create_pdf;
```
oferb: We use lowerCamelCase for member names, so `buttonCreatePdf`.
I would rename to `createPdfButton` to make it sound more like English.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,327 @@
...
+
+    /**
+     * This is where configurations of folders structure of Lawsuit Acitivity is been done.
+     * In case where directories don't exists, it creates them.
+     */
+    public void initDirStructure(){
+
+        // Main LawsuitActivity path:
+        // app_name/lawsuit_main_dir/
+        lawsuitMainPath = Paths.get(Environment.getExternalStorageDirectory().getPath(),getString(R.string.app_name),LAWSUIT_MAIN_DIR_NAME).toString();
```
oferb: Looks like `lawsuitMainPath` is not used. If you skip creating it, and create only `lawsuitTemplateDirPath`, will it create it recursively?

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,327 @@
...
+        if (!dirExists(lawsuitTemplateDirPath)){
+            File directory = new File(lawsuitTemplateDirPath);
+            directory.mkdirs();
+        }
+
+        // Template file's path:
+        try {
+            lawsuitTemplateFilePath = Paths.get(lawsuitTemplateDirPath, LAWSUIT_TEMPLATE_FILE_NAME).toString();
+        }
+        catch (Exception e){
```
oferb: So an exception is thrown by `Paths.get()` for non-existent files, but not for folders?

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,327 @@
...
+            res = true;
+        return res;
+    }
+
+    // Check permissions: TODO: Check if this should be done at an earlier activity.
+    public boolean isReadStoragePermissionGranted() {
+        if (Build.VERSION.SDK_INT >= 23) {
+            if (checkSelfPermission(Manifest.permission.READ_EXTERNAL_STORAGE)
+                    == PackageManager.PERMISSION_GRANTED) {
+                Log.w(TAG, "Permission is granted1");
```
oferb: Why the `1` in `granted1`?

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,327 @@
...
+                ActivityCompat.requestPermissions(this, new String[]{Manifest.permission.READ_EXTERNAL_STORAGE}, 3);
+                return false;
+            }
+        } else { //permission is automatically granted on sdk<23 upon installation
+            Log.v(TAG, "Permission is granted1");
+            return true;
+        }
+    }
+
+    public boolean isWriteStoragePermissionGranted() {
```
oferb: Looks like this doesn't need to be public.

oferb: The code for `isWriteStoragePermissionGranted()` is very similar to `isReadStoragePermissionGranted()`. Perhaps you could write a method that takes a parameter, and use only 1 method instead of 2?

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdf.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,327 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Paths;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.TimeZone;
+
+import noloan.R;
+
+public class LawsuitPdfActivity extends AppCompatActivity { // TODO: Add logs to class functions.
```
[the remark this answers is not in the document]
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,327 @@
...
+
+    private static final String TAG = "LawsuitPdfActivity";
+
+    // Folder names configurations:
+    private static final String LAWSUIT_MAIN_DIR_NAME = "Lawsuits";                  // Name of general lawsuit dir.
```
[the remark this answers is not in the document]
cotom972: Done. I did leave the comments for folder's meaning, let me know if you think it's not needed, and maybe makes the code more "messy".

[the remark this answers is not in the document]
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,327 @@
...
+    private static final String LAWSUIT_TEMPLATE_DIR_NAME = "Templates";             // Name of lawsuit template's dir.
+    private static final String LAWSUIT_OUTPUT_DIR_NAME = "Lawsuits PDFs";           // Name of resulted lawsuit PDFs output dir.
+    private static final String LAWSUIT_TEMPLATE_FILE_NAME = "LawsuitTemplate.xhtml";// Name of lawsuit's template file.
+    private static String lawsuitOutputPdfFileName = "";                             // Init when user confirms form.
+
+    // Folder paths (See: 'initDirStructure()' for structure between folders)
+    private static String lawsuitMainPath = "";
```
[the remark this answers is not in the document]
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: The thought was to have the date as close it can be to when user confirmed the form.
I changed createPdfLawsuitFileName() to be responsible just to generate the file name. The path for it is now generated inside the createPdf().

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: I wanted to avoid creating them recursively, so it will be easier to change the structure "between" them.

Changed it. Left the lawsuitMainPath to be as beginning of all other paths.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Changed this one to private. All other functions are set to protected. Should functions such as fillInTemplate() that is been used as part of createPdf() function, be set to private?

cotom972: I removed the isReadStoragePermissionGranted() - as if write permission is granted -> read permission as well. And use of this activity requires the write permission.

https://developer.android.com/training/data-storage/files

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Done. It isn't a class member now, and as a local variable of the createPdf() method.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: The exception is been thrown just for the missing template file.
Other folders, if don't exists, are just been created. I did not see this as an error, but rather as case of first use.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
oferb: Great.

oferb: I would set things to private, not protected. You don't plan to have this class derived from. If in the future it will happen, the dev will change private to protected where needed.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,316 @@
...
+        dateFormatter = new SimpleDateFormat(DATE_FORMAT);
+        dateFormatter.setTimeZone(TimeZone.getTimeZone(TIME_ZONE));
+
+        // Generate lawsuit PDF OnClick:
+        createPdfButton.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View view) {
+                try {
+                    createPdf();
+                } catch (IOException e) {
```
oferb: I think you can put this error inside `createPdf()`.
I believe you can then simplify this whole call using lambdas (Android Studio might suggest it as a lint warning).

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,316 @@
...
+        lawsuitMainPath = Paths.get(Environment.getExternalStorageDirectory().getPath(),getString(R.string.app_name),LAWSUIT_MAIN_DIR_NAME).toString();
+        if (!dirExists(lawsuitMainPath)){
+            File directory = new File(lawsuitMainPath);
+            directory.mkdirs();
+        }
+
+        // Templates dir path: 'lawsuitMainPath/LAWSUIT_TEMPLATE_DIR_NAME/'
+        // TODO: Currently template's file is been stored in user's device storage. Check to see if this should be part of the apk package.
+        lawsuitTemplateDirPath = Paths.get(lawsuitMainPath,LAWSUIT_TEMPLATE_DIR_NAME).toString();
+        if (!dirExists(lawsuitTemplateDirPath)){
```
oferb: Looks like you could simplify the code by extracting this to a method:
```
        if (!dirExists(lawsuitTemplateDirPath)){
            File directory = new File(lawsuitTemplateDirPath);
            directory.mkdirs();
        }
```

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,316 @@
...
+
+        return res;
+    }
+
+    /**
+     * Responsible for the output lawsuit PDF file name format.
+     * @return File name with ".pdf" extension.
+     */
+    protected String createPdfLawsuitFileName(){
+
```
oferb: You can just `return dateAndTimeFormatter.format(new Date()) + ".pdf";`

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,316 @@
...
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_lawsuit_pdf);
+        createPdfButton = (Button) findViewById(R.id.button_createPDF);
+
+        // Check Read/Write permissions:
+        isWriteStoragePermissionGranted();
```
oferb: Looks like you're not using the boolean return value..

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,316 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Paths;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.TimeZone;
+
+import noloan.R;
+
+public class LawsuitPdfActivity extends AppCompatActivity {
```
oferb: Can you please use 2-space indentation?

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,316 @@
...
+        res = res.replace("&lt;company2Fax&gt;", company2Fax);
+
+        return res;
+    }
+
+    /**
+     * Responsible for the output lawsuit PDF file name format.
+     * @return File name with ".pdf" extension.
+     */
+    protected String createPdfLawsuitFileName(){
```
oferb: When methods are short and don't do much, we usually call them `get`. `create` and `build` is for heavier stuff.
So I suggest renaming to `getPdfLawsuitFileName`.

oferb: Please follow this guide for naming:
https://github.com/hasadna/hasadna/blob/master/NOMENCLATURE.md

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,316 @@
...
+
+        String fileName = dateAndTimeFormatter.format(new Date()) + ".pdf";
+        return fileName;
+    }
+
+    /**
+     * This is where configurations of folders structure of Lawsuit Activity is been done.
```
oferb: When documenting methods - we use present simple:
https://www.grammarly.com/blog/simple-present/

Also, it's good to be short and to the point. For example, I would rewrite this as:
```
Create app folder structure.
```
Actually, now that I'm writing this, it doesn't seem to add information in addition to the method name, so we can probably just remove the comment altogether (unless I didn't notice some additional information in the comment).

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Moved it to createPdf().
About the lambdas, do we need to add Java 8 configurations?
https://developer.android.com/studio/write/java8-support

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Changed to void getWriteStoragePermission().

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Rearranged in general other comments as well, removed some unnecessary ones.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,234 @@
...
+import android.support.v4.app.ActivityCompat;
+import android.support.v7.app.AppCompatActivity;
+import android.os.Bundle;
+import android.util.Log;
+import android.view.View;
+import android.widget.Button;
+import android.widget.Toast;
+
+import com.itextpdf.licensekey.LicenseKey;
+import com.itextpdf.html2pdf.*;
```
oferb: Please don't use * on imports. If there are two * in the imports, given a class, we can't know for sure where it came from just by looking at the code. It also makes some analysis work harder to understand the dependencies :)

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,234 @@
...
+    private String lawsuitMainPath = "";
+    private String lawsuitTemplateDirPath = "";
+    private String lawsuitTemplateFilePath = "";
+    private String lawsuitOutputPath = "";
+
+    // Date formats
+    private static final String DATE_TIME_FORMAT = "dd-M-yyyy hh-mm-ss";
+    private static final String DATE_FORMAT = "dd-M-yyyy";
+    private static final String TIME_ZONE = "Asia/Jerusalem";
+    private SimpleDateFormat dateAndTimeFormatter;
```
oferb: If `dateAndTimeFormatter` is only dependent on `DATE_TIME_FORMAT` etc, then you can also make it `static final` (and also UPPER_SNAKE_CASE).

You can create a static method to set both format and timezone.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,234 @@
...
+        });
+
+    }
+
+    private void createPdf() {
+
+        // Activate iText license
+        // TODO: Find a better solution for reading the licensekey file (This was a just a "quick fix" to get it working).
+        InputStream licenseIs = getResources().openRawResource(R.raw.itextkey);
+        LicenseKey.loadLicenseFile(licenseIs);
```
oferb: For the most part, we try not to create local variables if not needed.
Exceptions are when the expression gets too complicated.

So here, I would use:
`LicenseKey.loadLicenseFile(getResources().openRawResource(R.raw.itextkey));`

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,234 @@
...
+                createPdf();
+            }
+        });
+
+    }
+
+    private void createPdf() {
+
+        // Activate iText license
+        // TODO: Find a better solution for reading the licensekey file (This was a just a "quick fix" to get it working).
```
oferb: Please describe what the fix would be (or at least, why is this not good as is).

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,234 @@
...
+        lawsuit = lawsuit.replace("&lt;company2Address&gt;", company2Address);
+        lawsuit = lawsuit.replace("&lt;company2Phone&gt;", company2Phone);
+        lawsuit = lawsuit.replace("&lt;company2Fax&gt;", company2Fax);
+
+        return lawsuit;
+    }
+
+    private String readTemplate() {
+
+        StringBuilder contentBuilder = new StringBuilder();
```
oferb: You can simplify using this:
https://stackoverflow.com/a/14169729

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,234 @@
...
+        // Output
+        lawsuitOutputPath = Paths.get(lawsuitMainPath, LAWSUIT_OUTPUT_DIR_NAME).toString();
+        initDir(lawsuitOutputPath);
+    }
+
+    /**
+     * Create dir if it doesn't exists.
+     * @param path Directory to look for.
+     */
+    private void initDir(String path) {
```
oferb: I'd rename this to `s/initDir/makeDir`. Everybody knows mkdir so it will ring a bell :)
Or, `createDir`, as besides creation, no other initialization is taking place, so that makes it more explicit.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,234 @@
...
+        File directory = new File(lawsuitTemplateDirPath);
+        if (directory.exists() && directory.isDirectory())
+            directory.mkdirs();
+    }
+
+    private void getWriteStoragePermission() {
+
+        if (Build.VERSION.SDK_INT >= 23)
+            if (!(checkSelfPermission(android.Manifest.permission.WRITE_EXTERNAL_STORAGE) == PackageManager.PERMISSION_GRANTED))
+                ActivityCompat.requestPermissions(this, new String[]{Manifest.permission.WRITE_EXTERNAL_STORAGE}, 2);
```
oferb: What happens if the user doesn't grant the permissions?

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Got you. Fixed it :)

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Ok thanks! Fixed it, but I had to initialize them on deceleration stage as they are finals. Correct?
About the time zones, created initTimeZones() for setting that. I'm not sure how much it values in term of code shortening, but it makes onCreate() much neater IMO. Is it worth it?

Removed `DATE_TIME_FORMAT` and `DATE_FORMAT` to be part of SimpleDateFormat construction. As part of trying to reduce unnecessary variables.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Changed it to be onClick as follows:
1. First time - request permission.
2. Denies - returns back to form.
3. Clicks again - Explain in a dialog why permission is need, and then asks for permission again.
    This will be continuous on every click attempt, unless user has checked "Do not show this 
    again" .
4. Click after "Do not show this again" - Only shows toast "Write storage permission needed."

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,295 @@
...
+
+  // Folder paths - initDirStructure()
+  private String lawsuitMainPath = "";
+  private String lawsuitTemplateDirPath = "";
+  private String lawsuitTemplateFilePath = "";
+  private String lawsuitOutputPath = "";
+
+  // Date formats
+  private static final String TIME_ZONE = "Asia/Jerusalem";
+  private static final SimpleDateFormat DATE_TIME_FORMATTER =
```
oferb: I can't comment on the old comment, so I'll comment here. I think what you did with the `SimpleDateFormat` is fine :)

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,295 @@
...
+      new SimpleDateFormat("dd-M-yyyy hh-mm-ss");
+  private static final SimpleDateFormat DATE_FORMATTER = new SimpleDateFormat("dd-M-yyyy");
+
+  // Initialise until App has client-side form
```
oferb: I would write this as a TODO:
`// TODO: Remove these once client-side form is done.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,295 @@
...
+
+  private String readTemplate() {
+
+    try {
+      return new String(
+          Files.readAllBytes(Paths.get(lawsuitTemplateFilePath)), StandardCharsets.UTF_8);
+    } catch (IOException e) {
+      Log.w(TAG, "Error reading lawsuit's template.");
+      e.printStackTrace();
+      return null;
```
oferb: If this returns null, you're going to get an NPE on `fillTemplate()`.
I think you should have a try-catch in `fillTemplate()` to handle that.
But, then this method just becomes `Files.readAllBytes(Paths.get(lawsuitTemplateFilePath)), StandardCharsets.UTF_8);`, which in that case, I suggest merging it with `fillTemplate()` :)

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,295 @@
...
+    DATE_FORMATTER.setTimeZone(TimeZone.getTimeZone(TIME_ZONE));
+  }
+
+  private void getWriteStoragePermission() {
+
+    if (Build.VERSION.SDK_INT >= 23) {
+
+      // Permission granted
+      if ((checkSelfPermission(android.Manifest.permission.WRITE_EXTERNAL_STORAGE)
+          == PackageManager.PERMISSION_GRANTED)) createPdf();
```
oferb: Please use `{}` on this `if`.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,295 @@
...
+    if (directory.exists() && directory.isDirectory()) directory.mkdirs();
+  }
+
+  private void initTimeZones() {
+
+    DATE_TIME_FORMATTER.setTimeZone(TimeZone.getTimeZone(TIME_ZONE));
+    DATE_FORMATTER.setTimeZone(TimeZone.getTimeZone(TIME_ZONE));
+  }
+
+  private void getWriteStoragePermission() {
```
oferb: Since this method calls `createPdf`, please make it clear from the method name.
Something like: `checkPermissionsThenCreatePdf()`.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,295 @@
...
+
+    if (Build.VERSION.SDK_INT >= 23) {
```
oferb: What if it's before 23?
That's more than 25% of devices.
https://developer.android.com/about/dashboards
Note that it's fine to limit, but please write some explanation about it.
AFAIU, the button doesn't do anything if it's before 23.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,295 @@
...
+    super.onCreate(savedInstanceState);
+    setContentView(R.layout.activity_lawsuit_pdf);
+    createPdfButton = (Button) findViewById(R.id.button_createPDF);
+
+    initDirStructure();
+    initTimeZones();
+
+    // Call createPdf() from callback
+    createPdfButton.setOnClickListener(
+        new View.OnClickListener() {
```
oferb: You can use lambdas here, I think:
`view -> getWriteStoragePermission()`.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: True. My mistake.

Fixed it. I assumed there's no need to check permissions for versions under 23.
By the way, I saw `minSdkVersion` is set to 28.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Changed it :)

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Done

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Fixed :)

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Changed it.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Thanks for commenting :)

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,264 @@
...
+
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_lawsuit_pdf);
+        createPdfButton = (Button) findViewById(R.id.button_createPDF);
+
+        initDirStructure();
+        initTimeZones();
+
+        // Call createPdf() from callback
+        createPdfButton.setOnClickListener(new View.OnClickListener() {
```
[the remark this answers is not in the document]
cotom972: Done.
Added these to app level _build.gradle_ file to support lambdas:

    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }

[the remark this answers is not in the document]
oferb: Great

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,281 @@
...
+          new String(
+              Files.readAllBytes(Paths.get(lawsuitTemplateFilePath)), StandardCharsets.UTF_8);
+    } catch (IOException e) {
+      Log.w(TAG, "Error reading lawsuit's template.");
+      e.printStackTrace();
+    }
+
+    // General form fields
+    lawsuit =
+        lawsuit.replace("&lt;claimCase&gt;", sentHaser ? claimCaseHaser : claimCaseSubscription);
```
oferb: You'll still get an NPE here. You can try it and see. Pass a bad `lawsuitTemplateFilePath`.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,281 @@
...
+    return lawsuit;
+  }
+
+  private String getPdfFilename() {
+
+    return DATE_TIME_FORMATTER.format(new Date()) + ".pdf";
+  }
+
+  private void initDirStructure() {
+
```
oferb: Can you please remove the newline at the beginning of methods? Thanks!

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,281 @@
...
+              .show();
+
+        } else {
+          ActivityCompat.requestPermissions(
+              this,
+              new String[] {Manifest.permission.WRITE_EXTERNAL_STORAGE},
+              STORAGE_PERMISSION_CODE);
+        }
+      }
+      // Devices running SDK <= 22, permissions granted on installation
```
oferb: Great comment!

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
oferb: We should change that. Can you ask Gal why that is on the channel?

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: :+1:

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: Fixed it as to our last conversation.

Write output / Read template exceptions are handled on `createPdf()`.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
cotom972: :+1:

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,275 @@
...
+    LicenseKey.loadLicenseFile(getResources().openRawResource(R.raw.itextkey));
+    try {
+      HtmlConverter.convertToPdf(
+          fillTemplate(),
+          new FileOutputStream(Paths.get(lawsuitOutputPath, getPdfFilename()).toString()));
+      Toast.makeText(LawsuitPdfActivity.this, "Lawsuit created!", Toast.LENGTH_LONG).show();
+
+      // Read template / write output
+    } catch (IOException e) {
+      Log.w(TAG, "Read template / Write file: " + e.getMessage());
```
oferb: You probably want to use a toast here too, to notify the user something went wrong.

projects/noloan2/app/src/main/java/hasadna/noloan2/LawsuitPdfActivity.java
```diff
@@ -0,0 +1,275 @@
...
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    setContentView(R.layout.activity_lawsuit_pdf);
+    createPdfButton = (Button) findViewById(R.id.button_createPDF);
+
+    initDirStructure();
+    setTimeZones();
+
+    createPdfButton.setOnClickListener((View v) -> checkPermissionsThenCreatePdf());
```
oferb: You don't even need the type and parenthesis I think:
`v -> checkPermissionsThenCreatePdf()`
Or,
`unused -> checkPermissionsThenCreatePdf()`

PR #121, opened 2019-02-28
Shared protoc - import proto

webdevelopland:
What's new:
- Ability to add .proto files, which contain import
- Ability to add .proto files from custom repos (not only startup-os)


projects/angular-proto-firestore/proto.json
```diff
@@ -3,7 +3,5 @@
   "projectProtoList": [
```
oferb: What are `projectProtoList`, `repoProtoList` and `shared-proto.json`?

common/PROTOC.md
```diff
@@ -3,6 +3,33 @@
...
+  ],
+  // List with paths to each .proto file inside the repo.
+  // The paths are relative to the repo
+  "repoProtoList": [
+    "common/repo/repo.proto"
+  ]
+}
+```
+
+There is `shared-proto.json`, which is list of paths to each .proto file inside the workspace.
```
elinabtl: Why do we have to declare the proto files in the workspace in advance?

common/PROTOC.md
```diff
@@ -3,6 +3,33 @@
...
+A project must contain `proto.json` file with paths to required files.  
+Example:
+```
+{ 
+  // Path to folder, where proto functions will be located.
+  // The path is relative to the project
+  "protoPath": "src/app/core/proto",
+  // List with paths to each .proto file inside the project.
+  // The paths are relative to the project
+  "projectProtoList": [
```
elinabtl: Why not have just 1 list of protos, whether in the project, outside the project but in the repo, or in a another repo (in the same workspace)?

common/PROTOC.md
webdevelopland: `shared-proto.json` contains list of proto, which is added to each project.

common/PROTOC.md
webdevelopland: Because it's more convinient.
`projectProtoList` - proto inside project (no need to set path to project)
`repoProtoList` - outside project, but inside repo (no need to set path to repo)
`shared-proto.json` - outside repo, but inside workspace (shared to all projects)

webdevelopland: But even can add one more list to 'proto.json`
`workspaceProtoList` - outside repo, but inside workspace (for specific project only)

elinabtl: What if we used only absolute paths always? E.g:
`startup-os/tools/reviewer/reviewer.proto`
`hasadna/projects/...`

oferb: What if we used only absolute paths always? E.g:
`startup-os/tools/reviewer/reviewer.proto`
`hasadna/projects/...`

common/PROTOC.md
oferb: Always added? 
What if a project doesn't need them?

webdevelopland: Well, it's optional. We can do not add the file to `shared-proto.json`.
Also, If you think there won't be a file, which is used in every project for sure, then we could remove the list.

common/PROTOC.md
webdevelopland: I don't know. In that way we don't have any pros, but got cons:
If a project change its location, we need to change each absolute proto path.
So I'd say several lists with relative paths is more practical way, but you have the last word.

oferb: Relative vs absolute is a philosophical question.
We discussed it and we'll go with absolute.

PR #133, opened 2019-03-10
Add template file to assets, general iText license key for Hasadna.

cotom972:


projects/noloan2/app/build.gradle
```diff
@@ -6,7 +6,7 @@ android {
     compileSdkVersion 28
     defaultConfig {
         applicationId "hasadna.noloan2"
-        minSdkVersion 14
+        minSdkVersion 26
```
cotom972: Paths.get() required min 26...

projects/noloan2/app/.gitignore
```diff
@@ -1 +1,3 @@
 /build
+/lib
+*.jar
```
cotom972: /lib/*.jar - ?

oferb: That's good. All jar files, all files under lib.. :)

projects/noloan2/app/build.gradle
oferb: In that case, we don't want Paths.get() since 26 leaves us with only 20% of users:
https://developer.android.com/about/dashboards

PR #134, opened 2019-03-10
Create startup-basics script

val-fed:


tools/startup/startup-basics-only.sh
```diff
@@ -0,0 +1,86 @@
...
+fi
+
+echo "Checking for curl installing"
+if [[ $(which curl; echo $?) == "1" ]]; then
+    echo "Installing curl"
+    sudo apt install curl
+    echo "curl is installed"
+fi
+
+echo "Checking for internet connection"
```
oferb: Perhaps this should be first? Assuming we have curl.

tools/startup/startup-basics-only.sh
```diff
@@ -0,0 +1,86 @@
...
+if [[ $(which google-chrome; echo $?) == "1" ]]; then
+  echo "Installing Chrome"
+  wget -q -O google-chrome-stable_current_amd64.deb https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
+  sudo dpkg -i google-chrome-stable_current_amd64.deb
+  sudo apt-get -f install
+fi
+
+# Install Bazel
+if [[ $(which bazel; echo $?) == "1" ]]; then
+    readonly BAZEL_VERSION=0.21.0
```
oferb: How about we put this const on the top?

tools/startup/startup-basics-only.sh
```diff
@@ -0,0 +1,86 @@
...
+# Creating base folder
+echo "Creating base folder"
+if [[ -d "$BASE_ABS_PATH" ]]; then
+    echo "$BASE_ABS_PATH folder already exists. Make sure you don't need $BASE_ABS_PATH folder, remove it and re-run the script."
+    exit 1
+fi
+mkdir ${BASE_ABS_PATH}
+
+# Set up StartupOS:
+git clone https://github.com/google/startup-os.git $DOWNLOAD_DIR_NAME/startup-os
```
oferb: What about bazel to PATH and bash completions?

tools/startup/startup-basics-only.sh
```diff
@@ -0,0 +1,86 @@
...
+bazel run //tools/reviewer/aa:aa_tool -- init --base_path $BASE_ABS_PATH
+cd ${BASE_ABS_PATH}/head/startup-os
+./test.sh
+
+# Set up .bashrc:
+cd ..
+echo "source `pwd`/startup-os/tools/reviewer/aa/aa_tool.sh" >> ~/.bashrc
+source ~/.bashrc
+
+echo "Please sign-in at https://web-login-startupos.firebaseapp.com and then check that ~/.aa_token exists"
```
oferb: Path is now `localhost:7000/login`

oferb: You can put this echo at the bottom.

tools/startup/startup-basics-only.sh
val-fed: This moved higher up.

tools/startup/startup-basics-only.sh
val-fed: Done.

tools/startup/startup-basics-only.sh
val-fed: Done. Please check this point one more time.

tools/startup/startup-basics-only.sh
val-fed: Done.

val-fed: Done. Please check my changes here. Is the path should be `https://localhost:7000/login`? 

When I run `aa` first time after running the script from `base` folder I get the invitation to visit `https://web-login-startupos.firebaseapp.com`. After log-in I have `.aa_token` file.

tools/startup/startup-basics-only.sh
oferb: What happens if internet is unavailable?
Does the `sudo apt install curl` exit the script?
If internet is unavailable, does it make sense to run `sudo apt install curl`?

tools/startup/startup-basics-only.sh
```diff
@@ -0,0 +1,87 @@
...
+fi
+
+# Install Bazel
+if [[ $(which bazel; echo $?) == "1" ]]; then
+ echo "Installing bazel"
+    sudo apt-get install pkg-config zip g++ zlib1g-dev unzip python
+    wget -O "bazel.${BAZEL_VERSION}.sh" "https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh"
+    chmod +x "bazel.${BAZEL_VERSION}.sh"
+    sudo ./bazel.${BAZEL_VERSION}.sh --user
+    export PATH="$PATH:$HOME/bin"
```
oferb: I think `export` won't change it for new terminals, only this terminal.

tools/startup/startup-basics-only.sh
oferb: `web-login-startupos.firebaseapp.com` is the old way. We will take down that server.
The new way is `localhost:7000/login`. Can you check it?

tools/startup/startup-basics-only.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+
+readonly DOWNLOAD_DIR_NAME="downloads_$(date +%Y%m%d_%H%M%S)"
+readonly DOWNLOAD_DIR_ABS_PATH="$(pwd)/$DOWNLOAD_DIR_NAME"
+readonly BASE_ABS_PATH=${HOME}/base
+readonly BAZEL_VERSION=0.21.0
```
oferb: We can already set this to the most recent Bazel version.

tools/startup/startup-basics-only.sh
val-fed: The script doesn't work without this line. It doesn't find `bazel` command in the next lines of the script.

tools/startup/startup-basics-only.sh
val-fed: Yes, you are right. It doesn't make sense to run `sudo apt install curl` if the internet is unavailable.
But we also can't check internet connection if `curl` isn't installed. 
I suggest in that case if `curl` isn't installed to print a message and exit the script:
```
Please ensure internet connection, install curl (sudo apt-get update && sudo apt-get install curl) and re-run the script.
```

tools/startup/startup-basics-only.sh
val-fed: Updated to `0.23.2`

tools/startup/startup-basics-only.sh
val-fed: I checked it. 
But `localhost:7000/login` isn't available while I don't run `aa` in the terminal. I was running just `aa`, without specifying a command.

tools/startup/startup-basics-only.sh
oferb: So I think maybe you need to add it *also* to bashrc.
Otherwise, if you open another terminal, can you call `bazel`?

PR #136, opened 2019-03-12
Add README.md for NoLoan2

cotom972:


projects/noloan2/README.md
```diff
@@ -0,0 +1,18 @@
+
+# Noloan2
+
+## Installation
+
+Download and add these 2 jars to `app/lib/`:
+- [IText 7 License Key Library](https://mvnrepository.com/artifact/com.itextpdf/itext-licensekey "IText 7 License Key Library")
+- [pdfCalligraph](https://mvnrepository.com/artifact/com.itextpdf/typography "typography")
+
+Change versions on app level `build.gradle` to fit the ones you downloaded:
```
oferb: It's the opposite way around - the person building the app doesn't know which versions to choose.
They need to look at `build.gradle` and get them from there.

So I would write this README as:
1. Go to `build.gradle` and see the versions for ...
2. Download them from these links and put them at ...

But, now that I think of it, Android Studio doesn't download them automatically once they're in `build.gradle`?

cotom972: Implementations for these 2 jars are as before, as downloaded files.

I didn't find a work around to implement them as others (Perhaps related to them been closed-source). Left it to a future task.

I "shrinked" the README file to how you proposed.

PR #150, opened 2019-03-25
add first stage task

lidorm:


projects/avodata/firstStage.py
```diff
@@ -0,0 +1,63 @@
...
+a = np.array([[-2,-2,4],[-1,-1,2]])
+M, A, e = find_c_and_a(a)
+print("The A matrix: ")
+print(A)
+print("The M natrix: ")
+print(M)
+print("The Error natrix: ")
+print(e)
+
+
```
oferb: Please add only 1-2 newlines at the end.

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,63 @@
...
+	return new_m, A, errors
+
+a = np.array([[-2,-2,4],[-1,-1,2]])
+M, A, e = find_c_and_a(a)
+print("The A matrix: ")
+print(A)
+print("The M natrix: ")
+print(M)
+print("The Error natrix: ")
```
oferb: `natrix->matrix`

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,63 @@
+import numpy as np
+import math
+
+def factorial(n):
```
oferb: Please use 2-space indentation using spaces, not tabs.

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,63 @@
...
+	return int(math.gamma(n + 1))
+
+
+def nChoosek(n, k):
```
oferb: We use the Google Style Guide:
https://github.com/google/styleguide/blob/gh-pages/pyguide.md

Please use snake_case for method names: `nChoosek->n_choose_k`

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,63 @@
...
+
+	num_of_rows_in_m = np.size(m,0)
+	c_matrix = np.zeros((num_of_rows_in_m, num_of_professions), dtype=np.float32)
+	for i in range(m.shape[0]):
+		c = np.dot(A, m[i].transpose())
+		c_matrix[i] = c
+
+	new_m = np.zeros((num_of_rows_in_m, num_of_professions), dtype=np.float32)
+	for i in range(c_matrix.shape[0]):
+		row_in_m = np.linalg.lstsq(A, c_matrix[i], None)[0] # TODO how can i got the error?
```
oferb: This is what we minimize: `min over M_i: ||A*M_i-C_i||`
`M, C` matrices, `M_i, C_i` vectors.

So `A*M_i-C_i` will be your error.

I see you already found the `errors`, so maybe this TODO can be removed?

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,63 @@
...
+
+def factorial(n):
+	return int(math.gamma(n + 1))
+
+
+def nChoosek(n, k):
+	return factorial(n) / (factorial(k) * factorial(n - k))
+
+
+def find_c_and_a(m):
```
oferb: Please document the shape of m, and the shape of the outputs.

projects/avodata/firstStage.py
yardenTal1: done

projects/avodata/firstStage.py
yardenTal1: done

projects/avodata/firstStage.py
yardenTal1: done

projects/avodata/firstStage.py
yardenTal1: done

projects/avodata/firstStage.py
yardenTal1: done

projects/avodata/firstStage.py
yardenTal1: done. I think.. this is what you mean?

projects/avodata/requirements.txt
```diff
@@ -0,0 +1,50 @@
+altgraph==0.10.2
```
oferb: This contains many libraries we're not using. I guess `freeze` wasn't a good idea.
I think you only need `numpy==1.8.0rc1`.

You can use `virtualenv`:
https://virtualenv.pypa.io/en/latest/
To create an isolated environment, and then see which dependencies you actually need.

BTW, how did you install all these libraries?

projects/avodata/firstStage.py
oferb: It's still tabs and not 2 spaces I believe.

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,55 @@
...
+	A = np.zeros(shape=(num_of_rows, num_of_professions), dtype=np.float32)
+	row = 0
+
+	for i in range(num_of_professions):
+		for j in range(i+1, num_of_professions):
+			A[row][i] = 1
+			A[row][j] = -1
+			row += 1
+
+	num_of_rows_in_m = np.size(m, 0)
```
oferb: Why do you use `np.size()` here and `m.shape()` above? If they do the same, it's better to be consistent.

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,55 @@
...
+
+def factorial(n):
+	return int(math.gamma(n + 1))
+
+
+def n_choose_k(n, k):
+	return factorial(n) / (factorial(k) * factorial(n - k))
+
+
+def find_c_and_a(m):
```
oferb: `find_c_and_a` is not what this method does.
I think it's a little confusing that we're combining the algorithm and the generation of synthetic data to check it.

I propose the following sketch:
1. `generate_synthetic_A` which receives `num_of_professions` and returns `A`
2. `generate_synthetic_c` which receives `m` and `i` and returns `c`.
They are called synthetic because normally they would be the result of user forms.
2. `solve` which receives `A` and `c` and returns a vector of `m`
3. `run_check` which receives `m`, for each vector generates synthetic data (you actually need to create `A` only once since it only depends on the structure of `m`), and then calls `solve`.

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,55 @@
+import numpy as np
```
oferb: Let's name this file `avodata.py` or `avodata_solver.py`

projects/avodata/firstStage.py
oferb: Yes, great. It will change a bit when we refactor.

projects/avodata/firstStage.py
yardenTal1: Can you help me with that? I tried before, and I'm not sure how ..

projects/avodata/firstStage.py
yardenTal1: How can i change it for all the file? its make me Balagan...

projects/avodata/firstStage.py
yardenTal1: I agree. I change :)

projects/avodata/firstStage.py
yardenTal1: For the first three points I think I understood and changed exactly as you wanted. For the last point, I created a new matrix M and a matrix of the same size with the errors, that's what you mean?

projects/avodata/firstStage.py
yardenTal1: Sorry, I think two changes are required, both for the file name and for the branch name. To the name of the branch made the changes according to the above - https://stackoverflow.com/questions/3866951/change-a-branch-name-in-a-git-repo
Is that ok???

and for the python file name, i will push with a new name after I've made all the changes.

oferb: No need to rename the branch.
Just the file. You just rename and commit it.

projects/avodata/firstStage.py
oferb: What IDE are you using?

projects/avodata/firstStage.py
oferb: As the output of `run_check`? Perfect.

projects/avodata/firstStage.py
oferb: Great :)

PR #157, opened 2019-03-29
Pull transitive dependencies of @startup_os so @startup_os's CI doesn't fail

vmax:


WORKSPACE
```diff
@@ -47,6 +47,9 @@ http_archive(
 # MARK: StartupOS end
 
 # StartupOS dependencies start
+load("@startup_os//third_party/maven:package-lock.bzl", maven_dependencies_startup_os="maven_dependencies")
+maven_dependencies_startup_os()
```
oferb: Please add a TODO:
`TODO: Check if order of `maven_dependencies()` and `maven_dependencies_startup_os()` in `WORKSPACE` dictates which version is used.`

PR #160, opened 2019-03-31
Add HeatMap to describe time from each point.

RamiNuraliyev:
HeatMap class input is stopId and time.
This class creates image that describe the arrival time of any place on the map to the Stop at that time by colors.
The brighter is the shortest time and the darkest is the longest.


projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,313 @@
...
+
+    graphics2D.dispose();
+    // save file
+    boolean img = false;
+    try {
+      img = ImageIO.write(image, "png", new File(imagePath.get()));
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+    if (img == false) System.out.println("can't create img");
```
val-fed: Type of `img` variable is boolean.  You can use: 
```
if (!img) {
  // do something
}
```
P.s. `img` isn't a good name for the boolean variable.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  public static void main(String[] args) {
+
+    Flags.parseCurrentPackage(args);
+    System.out.println(imagePath.get());
```
val-fed: It's ok to use "prints" during developing. But don't forget to delete it.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,313 @@
...
+      FileInputStream StopInput = null;
+
+      try {
+        StopInput =
+            new FileInputStream(new File("projects/opentrain/gtfs_pipeline/stops.protobin"));
+      } catch (FileNotFoundException e) {
+        e.printStackTrace();
+      }
+
+      while (true) { // from all stops class pick my stop from stopsTime *to use lon/lat
```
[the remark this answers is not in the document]
RamiNuraliyev: I don't think so because i choose stop from stopTime(i need DepartureTime and Stopsequence) and on this class i need stop from Stops class(i need Longitude and Latitude).

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,313 @@
...
+        } catch (IOException e) {
+          e.printStackTrace();
+        }
+      }
+    }
+  }
+
+  public void arrival_time_to_location() { // by stops.lat,stops.lon make matrix of colors
+
+    double max_lon = 0; // longitude
```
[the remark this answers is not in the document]
RamiNuraliyev: I don't see a difference because i don't use it so much,
why its better without a comment?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,313 @@
...
+    // location matrix *100 to make difference between points
+    heatMap = new int[(int) (max_lat * 100) + 1][(int) (max_lon * 100) + 1];
+
+    for (Stops_around stp : stopsAround) {
+
+      // arrival time of each stop from my stop by location
+      stp.i = (int) (stp.stop.getStopLat() * 100);
+      stp.j = (int) (stp.stop.getStopLon() * 100);
+      heatMap[stp.i][stp.j] = (int) (stp.arrivalTime);
+      // *100 to make difference between location
```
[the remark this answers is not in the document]
RamiNuraliyev: To 89, 90 i changed the stops location to increase the various.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,313 @@
...
+
+    public Stops_around(StopTime stopT, double arrivalT) {
+
+      arrivalTime = arrivalT;
+      Stop s;
+      FileInputStream StopInput = null;
+
+      try {
+        StopInput =
+            new FileInputStream(new File("projects/opentrain/gtfs_pipeline/stops.protobin"));
```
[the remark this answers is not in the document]
RamiNuraliyev: yes, vMax added it to the BUILD file.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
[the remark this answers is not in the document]
RamiNuraliyev: and on other methods,
I think that it's the best way to read the objects in the file until there is no more objects and there can be some comments and things that I don't need.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,313 @@
...
+
+  public class Stops_around // the stop and his location on arrival_time matrix
+   {
+
+    int i = 0; // index on heatMap
+    int j = 0;
+    Stop stop;
+    double arrivalTime;
+
+    public Stops_around(StopTime stopT, double arrivalT) {
```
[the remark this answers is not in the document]
RamiNuraliyev: stopT i can't change to stopTime because its a stop from stopTime not the stop time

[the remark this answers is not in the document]
val-fed: Sorry, maybe I lack context. What we want it's to replace variable names with more informative. It makes the code more readable and better to understand.
I don't see a problem in renaming `stopT` to `stopTime`. Could you please give more details or an example of why it isn't possible?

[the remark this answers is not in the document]
RamiNuraliyev: If you see variable that named stopTime you think that it's the time of the stop, but its just stop from the file stopTimes, there is other fields that i use like sequence, and for the time i have arrivalTime.
its feel incorrect, but i can change it.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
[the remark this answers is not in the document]
val-fed: I don't suggest removing all the comments in the code. Sometimes we can't avoid it. But if you can simplify/rename something to avoid comments please do it. It's better to understand what means variable/method/class just looking at the name than looking for comments for it. Sometimes code is changing but comments remain old and unactual. It can confuse.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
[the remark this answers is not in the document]
val-fed: In this case please move it above these lines. We are following a strategy when comments placed before the code in a separate line with a capital letter:
```
// A comment
some code
```

projects/opentrain/gtfs_pipeline/data_handler/DataHandler.java
```diff
@@ -56,9 +57,11 @@ private static boolean isServiceIdRelevantToday(CSVRecord record) {
     DateFormat dateFormat = new SimpleDateFormat("yyyyMMdd");
     int startDate = Integer.parseInt(record.get("start_date"));
     int endDate = Integer.parseInt(record.get("end_date"));
-    int currentDate = Integer.parseInt(dateFormat.format(new Date()));
+
+    // int currentDate = Integer.parseInt(dateFormat.format(new Date())); //Use if you want To read
```
oferb: Please put the comment before and make it explicit:
```
// Uncomment to read today's data:
// int currentDate = Integer.parseInt(dateFormat.format(new Date()));
// return (startDate <= currentDate && currentDate <= endDate && isRecordDayTrueOnDayOfToday);```

File: projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java [hunk 9]
```diff
@@ -0,0 +1,287 @@
...
+
+public class HeatMap {
+
+  private List<StopsAround> stopsAroundList = new ArrayList<>();
+
+  private int[][] heatMap;
+
+  @FlagDesc(
+      name = "image_path",
+      description = "Image path to create heatmap image,need to contain .../<FileName>.png")
```
oferb: `image,need->image. Needs`

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,287 @@
...
+
+    try (FileInputStream tripInputStream =
+            new FileInputStream(new File("projects/opentrain/gtfs_pipeline/trips.protobin"));
+        FileInputStream stopTimeInputStream =
+            new FileInputStream(new File("projects/opentrain/gtfs_pipeline/stopTime.protobin"))) {
+
+      // Test some trip stops heatMap
+      trip = Trip.parseDelimitedFrom(tripInputStream);
+      stopTime = StopTime.parseDelimitedFrom(stopTimeInputStream);
+      while (true) {
```
oferb: Why not:
`while (stopTime != null && trip != null)`

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,287 @@
...
+        if (stopTime == null | trip == null) {
+          break;
+        }
+
+        if (trip.getTripId().equals(stopTime.getTripId())
+            && Integer.parseInt(stopTime.getStopSequence()) == 1) {
+          HeatMap hm = new HeatMap(stopTime.getStopId(), stopTime.getDepartureTime());
```
oferb: Please use full names for variables: `hm->heatMap`. It's more readable that way.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,287 @@
...
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+
+  public void draw_map() {
```
oferb: We use lowerCamelCase for methods, so `draw_map->drawMap`

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,287 @@
...
+
+        float value = (float) (heatMap[i][j] - minData) / (float) (maxData - minData);
+
+        int r = (int) (((float) c2.getRed() - (float) c1.getRed()) * value + (float) c1.getRed());
+        int g =
+            (int) (((float) c2.getGreen() - (float) c1.getGreen()) * value + (float) c1.getGreen());
+        int b =
+            (int) (((float) c2.getBlue() - (float) c1.getBlue()) * value + (float) c1.getBlue());
+
+        Color c = new Color(r, g, b);
```
oferb: `c->color`. See other comment about naming.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,287 @@
...
+        graphics2D.setPaint(c);
+        graphics2D.fillRect(i, j, 2, 2);
+      }
+    }
+
+    graphics2D.dispose();
+    // Save file
+    boolean createImage = false;
+    try {
+      createImage = ImageIO.write(image, "png", new File(imagePath.get()));
```
oferb: It's better to have `imagePath.get()` passed to the method.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,287 @@
...
+      }
+    }
+  }
+
+  public HeatMap(int stopId, String time) throws IOException { // Time = HH:MM:ss
+
+    StopTime stopTime;
+    Trip trip;
+
+    try (FileInputStream tripsInputStream =
```
oferb: I think you could simplify the code if you load `trips.protobin` and `stopTime.protobin` into lists in memory.
They are small enough to fit. and it will also reduce I/O.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,287 @@
...
+            new FileInputStream(new File("projects/opentrain/gtfs_pipeline/trips.protobin"));
+        FileInputStream stopTimeInputStream =
+            new FileInputStream(new File("projects/opentrain/gtfs_pipeline/stopTime.protobin"));
```
oferb: Please make these paths `static final` consts.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
RamiNuraliyev: Do you mean that i don't need "new File" there?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,895 @@
...
+  // y = -1.28461208·10-6 x12 - 1.39031381·10-6 x1 x2
+  // - 3.786608865·10-7 x22 + 3.287492839·10-3 x1 + 9.395008917·10-4 x2 + 33.77106813
+  /* public double getLatitudeByXY ( int x, int y)
+  {
+    double latitude = -1.28461208 * Math.pow(10, -6) * Math.pow(x, 2) - 1.39031381 * Math.pow(10, -6) * x * y
+            - 3.786608865 * Math.pow(10, -7) * Math.pow(y, 2) + 3.287492839 * Math.pow(10, -3) * x
+            + 9.395008917 * Math.pow(10, -4) * y + 33.77106813;
+    return latitude;
+  }*/
+  private void arrivalTimeToLocationForMapNewFunction() {
```
oferb: I guess this is called NewFunction because the original is commented out.
Please rename this after removing the comments.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,895 @@
...
+    try (FileInputStream tripInputStream =
+            new FileInputStream(new File("projects/opentrain/gtfs_pipeline/trips.protobin"));
+        FileInputStream stopTimeInputStream =
+            new FileInputStream(new File("projects/opentrain/gtfs_pipeline/stopTime.protobin"))) {
+
+      // Test some trip stops heatMap
+      trip = Trip.parseDelimitedFrom(tripInputStream);
+      stopTime = StopTime.parseDelimitedFrom(stopTimeInputStream);
+
+      while (stopTime != null | trip != null) {
```
oferb: Why do you need to check `trip != null` every time if `trip` is not changed in the loop?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,895 @@
...
+      for (int j = 0; j < heatMap[0].length - 1; j++) {
+        if (heatMap[i][j] == 0) {
+          System.out.println("ERROR:" + "\n" + "heatmap matrix value should not be 0");
+        }
+
+        float value = (float) (heatMap[i][j] - minData) / (float) (maxData - minData);
+
+        int r =
+            (int)
+                (((float) color2.getRed() - (float) color1.getRed()) * value
```
oferb: I think it's enough to leave one float here, the first one.
float + int = float.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,895 @@
...
+        if (maxData < heatMap[i][j]) {
+          maxData = heatMap[i][j];
+        }
+        if (heatMap[i][j] != 0 && minData > heatMap[i][j]) {
+          minData = heatMap[i][j];
+        }
+      }
+    }
+
+    // Create image /** TODO: Try different size of image
```
oferb: `/**` seems unneeded.
I would also place the `TODO` on a separate line.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,895 @@
...
+                (((float) color2.getBlue() - (float) color1.getBlue()) * value
+                    + (float) color1.getBlue());
+
+        Color color = new Color(r, g, b);
+        graphics2D.setPaint(color);
+        graphics2D.fillRect(i, j, 2, 2);
+      }
+    }
+    // Draw "X" on stops location
+    for (StopsAround stp : stopsAroundList) {
```
oferb: Please use full names, i.e `stp->stopsAround`. It makes it easier to read.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,895 @@
...
+  public void drawMap(String imagePath) {
+
+    // color1 to color2
+    Color color1 = new Color(255, 208, 235);
+    Color color2 = new Color(126, 0, 67);
+
+    // Normalize data
+    double maxData = 0;
+    double minData = heatMap[10][10];
+    for (int i = 0; i < heatMap.length - 1; i++) {
```
oferb: Looks like you're skipping the last element. Is that intentional?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,895 @@
...
+      while (true) {
+        Stop tmpStop = Stop.parseDelimitedFrom(stopInputStream);
+        if (tmpStop == null) {
+          break;
+        }
+        stopsList.add(tmpStop);
+      }
+      while (true) {
+        StopTime tmpStopTime =
+            StopTime.newBuilder().build().parseDelimitedFrom(stopTimeInputStream);
```
oferb: Do you need `newBuilder().build()`? You don't for `Stop`.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,895 @@
...
+                stopsAroundList.add(new StopsAround(tripStops, arrivalTime));
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+
+  private void dataToList() throws IOException {
```
oferb: I'd rename this to `loadData` or `loadDataFromFiles`.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,895 @@
...
+          "Image destination path to create heatmap image,needs to contain .../<FileName>.png")
+  public static Flag<String> heatMapPath = Flag.create("");
+
+  public HeatMap(int stopId, String time) throws IOException { // Time = HH:MM:ss
+
+    dataToList();
+
+    // From all the stops
+    for (StopTime stopTime : stopTimeList) {
+      // Pick the stop that i want, That start some trip, At the chosen time
```
oferb: Let's rephrase to:
`Pick the stop that we want, that starts at some trip, at the given time`

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,895 @@
...
+  public static Flag<String> imagePath = Flag.create("");
+
+  @FlagDesc(
+      name = "heat_map_path",
+      description =
+          "Image destination path to create heatmap image,needs to contain .../<FileName>.png")
+  public static Flag<String> heatMapPath = Flag.create("");
+
+  public HeatMap(int stopId, String time) throws IOException { // Time = HH:MM:ss
+
```
oferb: We don't leave an empty line in the beginning of a method.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,895 @@
...
+              // Stops beside the chosen stop
+              if (tripStops.getTripId().equals(trip.getTripId())
+                  && tripStops.getStopId() != stopTime.getStopId()) {
+
+                // ArrivalTime HH:mm:ss to double seconds
+                String[] hhMMss = (tripStops.getArrivalTime().split(":"));
+                Double arrivalTime =
+                    Double.parseDouble(hhMMss[0]) * 3600
+                        + // H
```
oferb: Looks like `// H` should be before the `+`.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
RamiNuraliyev: If the trip file is empty i don't start the loop,
i added if for that case in the new HeatMap.java.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,500 @@
...
+    Flags.parseCurrentPackage(args);
+    StopTime stopTime;
+    Trip trip;
+
+    try (FileInputStream tripInputStream =
+            new FileInputStream(new File("projects/opentrain/gtfs_pipeline/trips.protobin"));
+        FileInputStream stopTimeInputStream =
+            new FileInputStream(new File("projects/opentrain/gtfs_pipeline/stopTime.protobin"))) {
+
+      // Test some trip stops heatMap
```
oferb: I'd rephrase to:
`// Take the first trip in file, for testing:`

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,500 @@
...
+    }
+  }
+
+  public static void main(String[] args) throws IOException {
+    // TODO:
+    //      Boarders of israel,do not paint sea area
+    //      Improve x,y pixel method by using more data
+
+    Flags.parseCurrentPackage(args);
+    StopTime stopTime;
```
oferb: Why define these outside the `try`?
If they're inside, you can set their values in the same line, e.g:
`StopTime stopTime = StopTime.parseDelimitedFrom(stopTimeInputStream);`

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,500 @@
...
+
+    try (FileInputStream tripInputStream =
+            new FileInputStream(new File("projects/opentrain/gtfs_pipeline/trips.protobin"));
+        FileInputStream stopTimeInputStream =
+            new FileInputStream(new File("projects/opentrain/gtfs_pipeline/stopTime.protobin"))) {
+
+      // Test some trip stops heatMap
+      trip = Trip.parseDelimitedFrom(tripInputStream);
+      stopTime = StopTime.parseDelimitedFrom(stopTimeInputStream);
+      if (trip == null) System.out.println("Error: trips file is empty");
```
oferb: Please use `{}` for if statements.
Also, I think you should exit in this case, since this is a failure.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,500 @@
...
+      if (stopTime == null) System.out.println("Error: stopTime file is empty");
+      while (stopTime != null) {
+        if (trip.getTripId().equals(stopTime.getTripId())
+            && Integer.parseInt(stopTime.getStopSequence()) == 1) {
```
oferb: Why do we require ` == 1`? Please document.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,500 @@
...
+
+          HeatMap heatMap = new HeatMap(stopTime.getStopId(), stopTime.getDepartureTime());
```
oferb: I don't understand - we're looping though all `stopTimes` and for each, creating a `HeatMap` instance and writing to the image? Shouldn't there be a single `HeatMap` instance?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,500 @@
...
+    // TODO:
+    //      Boarders of israel,do not paint sea area
+    //      Improve x,y pixel method by using more data
+
+    Flags.parseCurrentPackage(args);
+    StopTime stopTime;
+    Trip trip;
+
+    try (FileInputStream tripInputStream =
+            new FileInputStream(new File("projects/opentrain/gtfs_pipeline/trips.protobin"));
```
oferb: Please use the constant above - `tripsPath`.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,500 @@
...
+
+public class HeatMap {
+
+  private List<StopsAround> stopsAroundList = new ArrayList<>();
+
+  private double[][] heatMap;
+  private List<Stop> stopsList = new ArrayList<>();
+  private List<StopTime> stopTimeList = new ArrayList<>();
+  private List<Trip> tripsList = new ArrayList<>();
+  private static final String tripsPath = "projects/opentrain/gtfs_pipeline/trips.protobin";
```
oferb: `static finals` are constants, so should be `UPPER_CASE`. For example: `TRIPS_PATH`.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,500 @@
...
+  public static void main(String[] args) throws IOException {
+    // TODO:
+    //      Boarders of israel,do not paint sea area
+    //      Improve x,y pixel method by using more data
+
+    Flags.parseCurrentPackage(args);
+    StopTime stopTime;
+    Trip trip;
+
+    try (FileInputStream tripInputStream =
```
oferb: Looks like you're loading the files twice - once here and once inside `loadDataFromFiles()`.
Why not load them only once?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
RamiNuraliyev: I think it better like that because i allocate memory only once and use it like a pointer every time that i read from the file.
What do you think?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
RamiNuraliyev: No, we pick the first trip for testing, then we loop the stops until we get the stop that start this trip `(stopTime.getStopSequence()) == 1)`, `StopSequence `mean the index of this stop on the trip.
There will be only one `HeatMap` that get this stop id and the departure time from this stop.
When we find that stop, after the heat map creation, the loop stops (`break`).

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
RamiNuraliyev: I need to return to the start of the file,
i don't find good solution for that.
There is some solution with buffer but if the file is big it doesn't work.
Another problem is if i open the files in main i need to transfer them to `HeatMap` and then i need to make `try` on all the class so which function will close that.
If you know some way to read a file and return to its start point i will improve that.

oferb: What if you load it all into memory - into a list?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
oferb: Ohh ok :-)

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
RamiNuraliyev: If `loadDataFromFiles` is static i can use it without create `HeatMap` object but i can use it on main?

elinabtl: Sorry, can you clarify the question?

oferb: Can you clarify the question?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,549 @@
...
+  public int getXpixel(double longitude, double latitude) {
+    int xPixel =
+        (int)
+            ((31.84931551 * (Math.pow(longitude, 2)))
+                + (-155.5785894 * longitude * latitude)
+                + (134.2697945 * (Math.pow(latitude, 2)))
+                + (3427.951862 * longitude)
+                - (3846.07897 * latitude)
+                + 2512.552383);
+    /*#2(int)( -15980 + (69.73)*longitude + (405.4)*latitude);*/
```
oferb: Why is this commented out?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,549 @@
...
+  public int getYpixel(double longitude, double latitude) {
+    int yPixel =
+        (int)
+            ((67.87903886 * (Math.pow(longitude, 2)))
+                + (-59.89376472 * longitude * latitude)
+                + (11.68179254 * (Math.pow(latitude, 2)))
+                + (-2895.823625 * longitude)
+                + (1070.268543 * latitude)
+                + 38611.28426);
+    /*#2(int)(23100 + (-653.6)*longitude + (-58.28)*latitude);*/
```
oferb: Why is this commented out?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,549 @@
...
+
+    // Normalize data
+    double maxData = 0;
+    double minData = heatMap[10][10];
+    for (int i = 0; i < heatMap.length; i++) {
+      for (int j = 0; j < heatMap[0].length; j++) {
+        if (maxData < heatMap[i][j]) {
+          maxData = heatMap[i][j];
+        }
+        if (heatMap[i][j] != 0 && minData > heatMap[i][j]) {
```
oferb: Why do you need to check `heatMap[i][j] != 0`? Please explain in a comment.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,549 @@
...
+
+  // Draw only the heat map colors
+  public void drawHeatMap(String imagePath) {
+    // color1 to color2
+    Color color1 = new Color(255, 208, 235);
+    Color color2 = new Color(126, 0, 67);
+
+    // Normalize data
+    double maxData = 0;
+    double minData = heatMap[10][10];
```
oferb: Why `[10][10]` and not, say, `[0][0]`?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,549 @@
...
+
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+    if (!createImage) {
+      System.out.println("can't create image");
+    }
+  }
+
+  // Functions by different data and sites to get x pixel on the map:
```
oferb: `Functions by different data and sites to get x pixel on the map` - that sentence is not clear to me. Can you clarify?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,549 @@
...
+            + (-2895.823625 * longitude)
+            + (1070.268543 * latitude)
+            + 38611.28426);*/
+
+    return // yPixel;
+    yPixel + 53;
+  }
+
+  private void arrivalTimeToLocationForMap() {
+    heatMap = new double[770][918];
```
oferb: Please make these sizes `static finals`

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,549 @@
...
+    int y = getYpixel(startStop.getStopLat(), startStop.getStopLon());
+    int x = getXpixel(startStop.getStopLat(), startStop.getStopLon());
+    int startStopX = x;
+    int startStopY = y;
+    heatMap[x][y] = 0;
+
+    for (StopsAround stopAround : stopsAroundList) {
+
+      // Arrival time of each stop from the chosen stop by location
+      y = getYpixel(stopAround.stop.getStopLat(), stopAround.stop.getStopLon());
```
oferb: It's better to be consistent - first x, then y. So I would switch the line order here.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,549 @@
...
+      description = "Image path to map image. Needs to contain .../<FileName>.png")
+  public static Flag<String> imagePath = Flag.create("");
+
+  @FlagDesc(
+      name = "heat_map_path",
+      description =
+          "Image destination path to create heatmap image. Needs to contain .../<FileName>.png")
+  public static Flag<String> heatMapPath = Flag.create("");
+
+  public HeatMap() throws IOException {
```
oferb: I think a cleaner design would be to pass the lists to the `HeatMap` constructor, rather than load them inside. That way, someone could use `HeatMap` with different lists too.

projects/opentrain/gtfs_pipeline/heatmap/Data for Map
```diff
@@ -0,0 +1,50 @@
...
+31.516427, 35.084496 490
+32.114539, 35.174245 541
+32.118699, 34.213093 67 
+32.380066, 33.754649 22
+32.275572, 33.731573 0
+31.186971, 34.290645 0
+32.195446, 35.259177 609
+31.195970, 35.245844 609
+
+
```
val-fed: Do you need these newlines at the end of the file?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,540 @@
...
+    loadDataFromFiles();
+  }
+
+  public HeatMap(List<Stop> stopsList, List<StopTime> stopTimeList, List<Trip> tripsList) {
+    this.stopsList.addAll(stopsList);
+    this.stopTimeList.addAll(stopTimeList);
+    this.tripsList.addAll(tripsList);
+  }
+
+  public void createHeatMap(int stopId, String time) throws IOException {
```
val-fed: `IOException` is never thrown in the method. You can remove it here.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,540 @@
...
+
+    public StopsAround(StopTime stopTime, double arrivalTime) {
+
+      this.arrivalTime = arrivalTime;
+      setStop(stopTime);
+    }
+
+    private void setStop(StopTime stopTime) {
+      // From all stops class pick the chosen stop from stopsTime
+      // *to use longitude/latitude
```
val-fed: Looks like `*` sign is redundant here.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,540 @@
...
+
+        if (stopAround.getStopId() == stopTime.getStopId()) {
+          stop = stopAround;
+          break;
+        }
+      }
+    }
+  }
+
+  // arrivalTime To Location by stops.latitude,stops.longitude make matrix of colors
```
val-fed: Please use a capital letter for the first word in the comments.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,540 @@
...
+  private void arrivalTimeToLocation() {
```
val-fed: Could you please explain in the comment why this method isn't used?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,540 @@
...
+        }
+        if (heatMap[i][j] == 0) {
+          System.out.println("Check heatMap assignment");
+        }
+      }
+    }
+  }
+
+  // Draw only the heat map colors
+  public void drawHeatMap(String imagePath) {
```
val-fed: `imagePath` parameter is never used. Do we need it?

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,540 @@
...
+          System.out.println("i= " + i + "j=" + j);
+        }
+      }
+    }
+  }
+
+  private boolean inBorders(int x, int y) {
+    // Not in the Sea area
+    // y = -3.244698554·10-3 x2 - 5.987572519·10-1 x + 583.173052
+    if (y
```
val-fed: I think you can just return the result of the condition.
I mean:
```
    return (y
        > -3.373851571 * (Math.pow(10, -3)) * (Math.pow(x, 2))
        - 1.457345683 * (Math.pow(10, -1)) * x
        + 657.4869216);
```

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
```diff
@@ -0,0 +1,540 @@
...
+
+    graphics2D.drawImage(mapImage, 0, 0, null);
+
+    // Draw the map
+    for (int i = 0; i < heatMap.length; i++) {
+      for (int j = 0; j < heatMap[0].length; j++) {
+        if (heatMap[i][j] == 0) {
+          int x = getXpixel(startStop.getStopLat(), startStop.getStopLon());
+          int y = getYpixel(startStop.getStopLat(), startStop.getStopLon());
+          int startStopX = x;
```
val-fed: Looks like `startStopX` and `startStopY` variables are never used.

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
RamiNuraliyev: Yes, we can create two images once,
one that contains only the heat map colors
and the second contains the heat map on area map( Israel map).

projects/opentrain/gtfs_pipeline/heatmap/HeatMap.java
RamiNuraliyev: Previously i ignore the value of the start point and the distance of the points on the map from this point , and that caused color problems , so now i am calculating to each point the distance from the start point  as the initialize value and then search closer point, if there is one, and change the value.

PR #165, opened 2019-04-07
Avodata

yardenTal1:
Rename the branch to 'avodata'


projects/avodata/firstStage.py
```diff
@@ -0,0 +1,121 @@
...
+        for j in range(i + 1, num_of_professions):
+            A[row][i] = 1
+            A[row][j] = -1
+            row += 1
+
+    return A
+
+
+def generate_synthetic_c(m, i, a):
+    c = np.dot(a, m[i].transpose())
```
oferb: You can return directly, without an intermediate `c` variable.

Same for `solve`.

The name of the method already has the information of what is returned.

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,121 @@
...
+    for i in range(num_of_characteristics):
+        c = c_matrix[i]
+        row_in_m = solve(A, c)
+        row_in_m = (row_in_m - np.min(row_in_m)) / np.ptp(row_in_m)
+        m[i] = row_in_m
+    return m
+
+
+
+def run_check (m):
```
oferb: Extra space:
`run_check ( -> run_check(`

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,121 @@
...
+# print("The Error matrix: ")
+# print(E_matrix)
+
+# The professions are - driver, lawyer, programmer
+# The characteristics are - social status, working with people, analytical thinking
+c_matrix = [[-3, -3, -2],
+            [-1, 2, 3],
+            [-3, -3, -2]]
+m = build_m(3, c_matrix, 3)
+print(m)
```
oferb: Please add a newline at the end of the file.

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,121 @@
...
+        row_in_m = solve(A, c)
+        new_m[i] = row_in_m
+    Ax = np.dot(A, new_m.transpose())
+    errors = np.subtract(Ax, c_matrix.transpose())
+
+    return new_m, errors
+
+#
+# def find_c_and_a(m):
+#     # A function that receives a matrix m with size k*n, such that k is a characteristics of
```
oferb: If we leave commented code - we also explain why it's commented and when it should be uncommented.
Alternatively, if you don't need the code, you can delete it. It's always stored in git's history.

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,121 @@
+import numpy as np
+import math
+
```
oferb: Please give some general explanation of this file, and the different matrices and their dimensions.
It's basically what you wrote for `find_c_and_a` (now commented out).

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,121 @@
...
+    row_in_m = np.linalg.lstsq(A, c, None)[0]
+    return row_in_m
+
+def build_m(num_of_professions, c_matrix, num_of_characteristics):
+    A = generate_synthetic_a(num_of_professions)
+    m = np.zeros((num_of_characteristics, num_of_professions), dtype=np.float32)
+    for i in range(num_of_characteristics):
+        c = c_matrix[i]
+        row_in_m = solve(A, c)
+        row_in_m = (row_in_m - np.min(row_in_m)) / np.ptp(row_in_m)
```
oferb: `ptp` is just `max` in this case, since you subtract the minimum value, so the smallest value is always 0.
So I would just replace it with `max`.
Also, please comment before:
`// We normalize to the [0,1] range:

projects/avodata/firstStage.py
yardenTal1: done

projects/avodata/firstStage.py
yardenTal1: done

projects/avodata/firstStage.py
yardenTal1: Cool. good to know :)

projects/avodata/firstStage.py
yardenTal1: yes :)

projects/avodata/firstStage.py
yardenTal1: If I just change to max the result is not the same. Because the smallest value is 0 just after all this action actually. So max and ptp are not the same here. 
It's possible with two rows, and i did it... 
I hope its good like this.

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,79 @@
+import numpy as np
+import math
+
+# A program that receives a score matrix (c_matrix) for characteristics that characterize certain subjects.
+# And produces a matrix M with size k*n (such that k is a characteristics of
+# professions and n is the number of professions that are compared),
+# that shows the ratio of the properties between the different professions
```
oferb: I'd rephrase `ratio->relative value`. Ratio is usually for `a/b`.

projects/avodata/firstStage.py
```diff
@@ -0,0 +1,79 @@
...
+# The program using the List Square algorithm, the vector c with size k*1, and a
```
oferb: `Square->Squares`

projects/avodata/firstStage.py
yardenTal1: done

projects/avodata/firstStage.py
yardenTal1: done

projects/avodata/firstStage.py
```diff
@@ -1,11 +1,13 @@
...
+import pandas as pd
+
 
 # A program that receives a score matrix (c_matrix) for characteristics that characterize certain subjects.
 # And produces a matrix M with size k*n (such that k is a characteristics of
 # professions and n is the number of professions that are compared),
-# that shows the ratio of the properties between the different professions
-# The program using the List Square algorithm, the vector c with size k*1, and a
+# that shows the relative value of the properties between the different professions
+# The program using the List Squares algorithm, the vector c with size k*1, and a
```
oferb: `List->Least`

projects/avodata/firstStage.py
```diff
@@ -72,8 +79,11 @@ def run_check(m):
 
 # The professions are - driver, lawyer, programmer
 # The characteristics are - social status, working with people, analytical thinking
-c_matrix = [[-3, -3, -2],
-            [-1, 2, 3],
-            [-3, -3, -2]]
+# c_matrix = [[-3, -3, -2],
```
oferb: Please delete things you don't need :)

projects/avodata/firstStage.py
```diff
@@ -72,8 +79,11 @@ def run_check(m):
...
+#             [-1, 2, 3],
+#             [-3, -3, -2]]
+
+c_matrix = create_c_matrix_from_csv('check.csv')
```
oferb: How about you add a `check.csv` example so people know how to run the code?
Also, would be good to explain the structure of `check.csv`.

projects/avodata/firstStage.py
yardenTal1: done

projects/avodata/firstStage.py
yardenTal1: done

projects/avodata/firstStage.py
yardenTal1: What you mean in the first sentence?

yardenTal1:
Merged avodata python file

PR #170, opened 2019-04-11
Filter stories draft

webdevelopland:


projects/angular/storyteller/src/app/routes/feed/feed.component.scss
```diff
@@ -1,6 +1,7 @@
 .page-container {
   >.align-container {
     >.feed {
+      width: 930px;
```
webdevelopland: Fixed width of feed block, to avoid its disappearing, when it doesn't contain any story.

projects/angular/storyteller/src/app/routes/feed/feed.component.ts
```diff
@@ -66,11 +68,9 @@ export class FeedComponent {
   // Filters stories by project id
   loadProjectStories(projectId: string): void {
     if (projectId !== 'all') {
-      this.easyStories = this.easyStories.filter(story => {
-        return story.project === projectId;
-      });
+      this.filteredEasyStories = this.storyService.getFilteredStories(this.easyStories, projectId);
```
webdevelopland: I decided, that that solution is simpler.
But if you wonder, how to load stories on each filter, I can show you an implementation.

projects/angular/storyteller/src/app/routes/user-info/user-info.component.ts
```diff
@@ -62,11 +64,9 @@ export class UserInfoComponent {
   // Filters stories by project id
   loadProjectStories(projectId: string): void {
     if (projectId !== 'all') {
-      this.easyStories = this.easyStories.filter(story => {
-        return story.project === projectId;
-      });
+      this.filteredEasyStories = this.storyService.getFilteredStories(this.easyStories, projectId);
     } else {
-      this.loadReviewerConfig(this.email);
+      this.filteredEasyStories = this.easyStories.slice();
```
webdevelopland: `slice` copies an array, without copying a object link

projects/angular/storyteller/src/styles/styles.scss
```diff
@@ -4,6 +4,8 @@
 
 html,
 body {
+  width: 100vw;
+  overflow-x: hidden;
```
webdevelopland: To avoid changing screen width by scroll

PR #185, opened 2019-04-26
avodataJavaCode

lidorm:
using least squares alg in java


ls-alg/check.csv
```diff
@@ -0,0 +1,4 @@
+d,l,p
```
oferb: Please put this in `projects/avodata`

ls-alg/main/Main.java
```diff
@@ -0,0 +1,162 @@
...
+ * To build:
+ * bazel build //ls-alg/main:ls-alg_tool
+ * To run:
+ * bazel-bin/ls-alg/main/ls-alg_tool
+*/
+
+
+
+
+public class Main {
```
oferb: Please run `./fix-formatting.sh` to fix the formatting.

ls-alg/main/Main.java
```diff
@@ -0,0 +1,162 @@
...
+	public static RealMatrix readMatrixAndCreate() {
+		
+		 Scanner scanner;
+		 StringBuilder b = new StringBuilder();
+		 RealMatrix rMatrix = new Array2DRowRealMatrix(3,3);
+		 Path currentRelativePath = Paths.get("");
+		 String myPath = currentRelativePath.toAbsolutePath().toString();
+		 
+		try {
+			scanner = new Scanner(new File(myPath +"/ls-alg/check.csv"));
```
oferb: Please make `"/ls-alg/check.csv"` a `static final`.

ls-alg/main/Main.java
```diff
@@ -0,0 +1,162 @@
...
+		//the final matrix after build
+		return rMatrix;
+	}
+	
+	
+	
+	public static void main(String[] args) {
+		
+		
+		//read matrix from file and build
```
oferb: Comments should start with a capital letter and a space, so for example here:
`//read`
would change to:
`// Read`

ls-alg/main/Main.java
```diff
@@ -0,0 +1,162 @@
...
+			e.printStackTrace();
+		}
+		//the final matrix after build
+		return rMatrix;
+	}
+	
+	
+	
+	public static void main(String[] args) {
+
```
oferb: Please avoid newlines at the beginning of methods.

ls-alg/main/Main.java
```diff
@@ -0,0 +1,162 @@
...
+import org.apache.commons.math3.linear.RealMatrix;
+import org.apache.commons.math3.linear.RealVector;
+import org.apache.commons.math3.linear.SingularValueDecomposition;
+import org.apache.commons.math3.util.CombinatoricsUtils;
+
+
+
+
+
+/**
```
oferb: It's great that you added this. Please also describe the whole class.

ls-alg/main/Main.java
```diff
@@ -0,0 +1,162 @@
+package main;
```
oferb: Let's rename this file `AvodataSolverTool.java` since it's a runnable (hence tool) and `AvodataSolver` seems like a good name, unless you can think of a better one.

oferb: I don't think we need another folder for `main`. It can all be in the same folder.

ls-alg/main/BUILD
```diff
@@ -0,0 +1,21 @@
+
```
oferb: Let's delete the old code.. :-)

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.commons.math3.analysis.function.Abs;
+import org.apache.commons.math3.linear.Array2DRowRealMatrix;
+import org.apache.commons.math3.linear.DecompositionSolver;
+import org.apache.commons.math3.linear.MatrixUtils;
+import org.apache.commons.math3.linear.RealMatrix;
+import org.apache.commons.math3.linear.RealVector;
+import org.apache.commons.math3.linear.SingularValueDecomposition;
+import org.apache.commons.math3.util.CombinatoricsUtils;
+
+/** To build: bazel build //projects/avodata:ls-alg_tool To run: bazel-bin/projects/avodata/ls-alg_tool */
```
oferb: Please use a multiline comment. Starting with `/*` prevents the formatter from formatting the comment (which is good since you want to preserve the newlines).
```
/* This class reads matrix from csv file and calculates profession attributes using least squares.
 * To build:
 * bazel build //projects/avodata:ls-alg_tool
 * To run:
 * bazel-bin/projects/avodata/ls-alg_tool
 */
```

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,142 @@
...
+/** This class reads matrix from csv file, ignores the parmeters and calculate 
+/** using least squares */
+public class AvodataSolverTool {
+
+  static final String fpath = "/projects/avodata/check.csv";
```
oferb: static finals are constants, so are `UPPER_CASE`.
Also, please use full names (e.g `FILE_PATH`).
I would actually change this to `CSV_PATH` to be more descriptive.

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,142 @@
...
+  public static double factorial(int n) {
+    return CombinatoricsUtils.factorialDouble(n);
+  }
+
+  public static double nChooseK(int n, int k) {
+    return CombinatoricsUtils.binomialCoefficientDouble(n, k);
+  }
+
+  public static double[][] generateSyntheticA(int numOfProfessions) {
+
```
oferb: Please avoid newlines at beginning of methods.

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  public static double[][] generateSyntethicC(RealMatrix m, int i, RealMatrix a) {
+    RealMatrix c = a.multiply(m.getRowMatrix(i).transpose());
+    return c.getData();
+  }
+
+  public static RealVector solve(RealMatrix a, RealVector c) {
+
+    SingularValueDecomposition s = new SingularValueDecomposition(a);
```
oferb: Did you consider using `org.apache.commons.math3.fitting.leastsquares`?
https://commons.apache.org/proper/commons-math/userguide/leastsquares.html

http://commons.apache.org/proper/commons-math/javadocs/api-3.6.1/org/apache/commons/math3/fitting/leastsquares/LeastSquaresProblem.html

BTW, where did you come up with the `SingularValueDecomposition` solution?

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,142 @@
...
+
+  public static double factorial(int n) {
+    return CombinatoricsUtils.factorialDouble(n);
+  }
+
+  public static double nChooseK(int n, int k) {
+    return CombinatoricsUtils.binomialCoefficientDouble(n, k);
+  }
+
+  public static double[][] generateSyntheticA(int numOfProfessions) {
```
oferb: I see sometimes you're returning `double[][]` and sometimes `RealMatrix`. It's better to be consistent.

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,142 @@
...
+      rowInM.mapAddToSelf(min);
+      double max = av.value(rowInM.getMaxValue());
+      rowInM.mapDivideToSelf(max);
+      m.setRowVector(i, rowInM);
+    }
+
+    return m.getData();
+  }
+
+  //  We read the matrix from csv file.
```
oferb: `//  We->// We`

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,142 @@
...
+    Scanner scanner;
+    StringBuilder b = new StringBuilder();
+    RealMatrix rMatrix = new Array2DRowRealMatrix(3, 3);
+    Path currentRelativePath = Paths.get("");
+    String myPath = currentRelativePath.toAbsolutePath().toString();
+
+    try {
+      scanner = new Scanner(new File(myPath + fpath));
+      scanner.useDelimiter(",");
+      // build one string of all the matrix
```
oferb: Please start comments with a capital letter:
`// build->// Build`

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,142 @@
...
+      int numOfProfessions, RealMatrix cMatrix, int numOfCharacteristics) {
+    RealMatrix a = new Array2DRowRealMatrix();
+    a = MatrixUtils.createRealMatrix(generateSyntheticA(numOfProfessions));
+    RealMatrix m = new Array2DRowRealMatrix(numOfCharacteristics, numOfProfessions);
+
+    for (int i = 0; i < numOfCharacteristics; i++) {
+
+      RealVector c = cMatrix.getRowVector(i);
+      RealVector rowInM = solve(a, c);
+      Abs av = new Abs();
```
oferb: Please use descriptive names for variables, unless it's super-common (like `i` for loops).
In this case, I can guess this means `absolute value`, but I'm not sure, and it's easier to read when it's a full word or words.

projects/avodata/AvodataSolverTool.java
lidorm: Ok, changed every method to return RealMatrix because it's the interface and more abstract.

projects/avodata/AvodataSolverTool.java
lidorm: I didn't very well the example they have there when i found apache common library so i tried to find 
simpler solution, I found the  **SingularValueDecomposition** solution on stack overflow.
I can try use the **org.apache.commons.math3.fitting.leastsquares** if you think its better.

projects/avodata/AvodataSolverTool.java
oferb: Cool

projects/avodata/AvodataSolverTool.java
oferb: Let's try to use the leastsquares again and see where the problem is.

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,143 @@
...
+import org.apache.commons.math3.analysis.function.Abs;
+import org.apache.commons.math3.linear.Array2DRowRealMatrix;
+import org.apache.commons.math3.linear.DecompositionSolver;
+import org.apache.commons.math3.linear.MatrixUtils;
+import org.apache.commons.math3.linear.RealMatrix;
+import org.apache.commons.math3.linear.RealVector;
+import org.apache.commons.math3.linear.SingularValueDecomposition;
+import org.apache.commons.math3.util.CombinatoricsUtils;
+
+/* This class reads matrix from csv file, ignores the parmeters and calculate
```
oferb: What do you mean by `ignores the parmeters`?

oferb: Please note my earlier comment, I rephrased a bit.

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,143 @@
...
+ * using least squares
+ * To build:
+ * bazel build //projects/avodata:ls-alg_tool
```
oferb: Let's call it `avodata-solver_tool`

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,143 @@
...
+      for (int j = i + 1; j < numOfProfessions; j++) {
+        matrix.addToEntry(row, i, 1);
+        matrix.addToEntry(row, j, -1);
+        row++;
+      }
+    return matrix;
+  }
+
+  public static RealMatrix generateSyntethicC(RealMatrix m, int i, RealMatrix a) {
+    RealMatrix c = a.multiply(m.getRowMatrix(i).transpose());
```
oferb: You don't need an intermediate variable. You can just return directly:
`return a.multiply(m.getRowMatrix(i).transpose());`

projects/avodata/AvodataSolverTool.java
oferb: Can you link to the StackOverflow you used?

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,143 @@
...
+
+  public static RealVector solve(RealMatrix a, RealVector c) {
+    SingularValueDecomposition s = new SingularValueDecomposition(a);
+    DecompositionSolver d = s.getSolver();
+    return d.solve(c);
+  }
+
+  public static RealMatrix buildCheck(
+      int numOfProfessions, RealMatrix cMatrix, int numOfCharacteristics) {
+    RealMatrix a = new Array2DRowRealMatrix();
```
oferb: You're instantiating and then assigning, so you don't really need the instantiation (`new`).

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,143 @@
...
+    RealMatrix m = new Array2DRowRealMatrix(numOfCharacteristics, numOfProfessions);
+
+    for (int i = 0; i < numOfCharacteristics; i++) {
+
+      RealVector c = cMatrix.getRowVector(i);
+      RealVector rowInM = solve(a, c);
+      Abs absoluteValue = new Abs();
+      double min = absoluteValue.value(rowInM.getMinValue());
+      rowInM.mapAddToSelf(min);
+      double max = absoluteValue.value(rowInM.getMaxValue());
```
oferb: What if this is 0?
It means all values are 0 I think, but in this case, I guess it's fine to return all values, so I would check if it's a 0 and if so, not divide.

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,143 @@
...
+      while (scanner.hasNext()) {
+        b.append(scanner.next() + " ");
+      }
+      // Build string array,every string is a vector of the matrix
+      String[] s = b.toString().split("\r\n");
+      int row = 0;
+      for (String string : s) {
+
+        String[] vectorToSet = string.split(" ");
+        // Ignorance of alphabet chars vector
```
oferb: Why do we have an alphabet chars vector in the csv, if we just want to ignore it? :)

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,143 @@
...
+  // We ignore the alphabet chars and build the matrix without them.
+  public static RealMatrix readMatrixAndCreate() {
+    Scanner scanner;
+    StringBuilder b = new StringBuilder();
+    RealMatrix rMatrix = new Array2DRowRealMatrix(3, 3);
+    Path currentRelativePath = Paths.get("");
+    String myPath = currentRelativePath.toAbsolutePath().toString();
+
+    try {
+      scanner = new Scanner(new File(myPath + CSV_PATH));
```
oferb: How about this?
`scanner = new Scanner(new File(Paths.get("").toAbsolutePath() + CSV_PATH));`
It's shorter, and I think also clearer.

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,143 @@
...
+
+    return m;
+  }
+
+  // We read the matrix from csv file.
+  // We ignore the alphabet chars and build the matrix without them.
+  public static RealMatrix readMatrixAndCreate() {
+    Scanner scanner;
+    StringBuilder b = new StringBuilder();
+    RealMatrix rMatrix = new Array2DRowRealMatrix(3, 3);
```
oferb: Is `r` for `real`? I think you can just call is `matrix` in that case.

projects/avodata/AvodataSolverTool.java
```diff
@@ -0,0 +1,143 @@
...
+      rowInM.mapDivideToSelf(max);
+      m.setRowVector(i, rowInM);
+    }
+
+    return m;
+  }
+
+  // We read the matrix from csv file.
+  // We ignore the alphabet chars and build the matrix without them.
+  public static RealMatrix readMatrixAndCreate() {
```
oferb: Please add a TODO to redo this method using a CSV reader, for example here:
https://github.com/hasadna/hasadna/blob/master/projects/opentrain/gtfs_pipeline/data_handler/DataHandler.java

projects/avodata/AvodataSolverTool.java
lidorm: Sure

https://stackoverflow.com/questions/5010682/numpy-linalg-lstsq-equivalent-in-apache-commons-math-java-library

projects/avodata/AvodataSolverTool.java
lidorm: Got it,done.

projects/avodata/AvodataSolverTool.java
```diff
@@ -81,9 +81,21 @@ public static RealMatrix buildCheck(
...
+  
   // We read the matrix from csv file.
   // We ignore the alphabet chars and build the matrix without them.
+  // For example:
+  /* We get this matrix from the CSV file.
+   * | a | b | c |
+   * | 1 | 2 | 3 |
+   * | 4 | 5 | 6 |
+   * | 7 | 8 | 9 |
+   * Whem we build the matrix we don't want the parameteres a,b,c.
```
oferb: `Whem->When`

PR #192, opened 2019-05-13
Hasadna local - Added file to track users' data

a3l5o7n9:
Added reviewer_config.prototxt file.


reviewer_config.prototxt
```diff
@@ -0,0 +1,93 @@
+display_name: "StartupOS"
+repo {
+  id: "startup-os"
+  url: "https://github.com/google/startup-os"
+}
+repo {
+  id: "hasadna"
+  url: "https://github.com/hasadna/hasadna"
+}
+project {
```
val-fed: I'd remove all projects because it relates to StartupOS. Instead of it add:
// TODO: Add hasadna's projects

reviewer_config.prototxt
```diff
@@ -0,0 +1,93 @@
...
+}
+user {
+  id: "oferb"
+  project_id: "google_oss_examples"
+  project_id: "reviewer"
+  top_contribution {
+    description: "Team leader for StartupOS project (Google OSS examples, Reviewer)"
+    url: "https://github.com/google/startup-os"
+  }
+  top_contribution {
```
val-fed: I'd remove all `top contribution` for all users exclude you. They will edit the file and add actual data by your own.

oferb: We can fix it later. For now it will help with testing.

reviewer_config.prototxt
oferb: Good idea.

PR #196, opened 2019-06-05
Renaming the 2 and reorganizing the files

gulilo:


projects/noloan/app/src/main/java/hasadna/noloan/Firestore/FirestoreElement.java
```diff
@@ -1,5 +1,6 @@
-package hasadna.noloan2;
+package hasadna.noloan.Firestore;
```
oferb: Package names (and corresponding folder names) should be lowercase (also for `Lawsuit`).

PR #197, opened 2019-06-06
Storyteller - mobile

webdevelopland:
Responsive design for mobile and tablet.


projects/angular/storyteller/src/app/shared/feed/feed.component.ts
```diff
@@ -57,6 +59,7 @@ export class FeedComponent implements OnInit, OnDestroy {
     this.displayStories();
   }
 
+  // Puts limited about of stories to a variable, which is used in html template
```
oferb: What's "limited about of stories"?

webdevelopland: amount*

PR #209, opened 2019-08-07
Storyteller - login & new mocks

webdevelopland:
What's new:
- Implementation of new design mocks
- Ability to log-in and log-out. Stories are visible to authorized users only.
- New proto (without projectId. Also small rearrangements)
- New names for projects:
  stotyteller-web -> stotyteller
  stotyteller-tracker -> old-tracker
- Removing projectId from old-tracker too.

What's still under developing:
- "Time tracking" page
- Tracker time in header and related popups


projects/angular/storyteller/src/app/shared/header/header.component.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  constructor(
+    public headerService: HeaderService,
+    public authService: AuthService,
+    public loadingService: LoadingService,
+    private router: Router,
+  ) {
+    this.router.events.subscribe(event => {
+      if (event instanceof NavigationStart) {
+        // Convert '/my-page?param=data' -> '/my-page'
```
oferb: LOL

webdevelopland: :)

PR #210, opened 2019-08-15
noloan Admin app

gulilo:


projects/noloan/admin/app/src/main/java/hasadna/noloanadmin/MainActivity.java
```diff
@@ -0,0 +1,13 @@
+package hasadna.noloanadmin;
```
oferb: How about `hasadna.noloan.admin.app`?

PR #219, opened 2019-09-23
Add: User id validation in form fragment, labels for editTexts

cotom972:


oferb:
Great improvement! Thanks.

PR #223, opened 2019-09-27
Step 1.5, deploy the project to Firebase hosting 

mohammad-suliman:
This PR deploys the project to Firebase hosting by adding the required files to this project. Note also that the ID attribute of the process button was removed as agreed with Ofer.


projects/sensory_interface/public/index.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
```
oferb: Why do you need `index.html` twice?
If you need it in `public`, you can remove it from the parent folder. Same for `app.js`.

projects/sensory_interface/index.html
```diff
@@ -9,9 +9,18 @@
 <body>
     <textarea id="textInput">1	2	3
 4	5	6</textarea>
-    <button id=" processButton" onclick="processData()">process </button>
+    <button onclick="processData()">process </button>
     <br>
     <div id="tableContainer"></div>
+    <!-- The core Firebase JS SDK is always required and must be listed first -->
```
oferb: Right now you're using only Firebase hosting, and I don't think you need the scripts for that.
Try removing them and deploying and see if it still works.

projects/sensory_interface/index.html
```diff
@@ -9,9 +9,18 @@
 <body>
     <textarea id="textInput">1	2	3
 4	5	6</textarea>
-    <button id=" processButton" onclick="processData()">process </button>
```
oferb: Good. ID is only needed if you want to obtain the element by ID, or reference it by ID from the CSS.

projects/sensory_interface/index.html
mohammad-suliman: it works without the scripts as you suggested. Removing...

oferb: Great :)

PR #227, opened 2019-10-06
Sensory interface: step 3: improve touch and sound capability 

mohammad-suliman:
This PR implements the step 3 of the Sensory interface project.


projects/sensory_interface/public/app.js
```diff
@@ -1,12 +1,20 @@
 // initialize Audio context on page load.
 let AudioContext = window.webkitAudioContext || window.AudioContext;
 let audioContext = new AudioContext();
+let oscillator = null;
+// This variable is not used currently.
+let source = null;
+// A variable which holds the current table cell under the touch point.
+let currentCellUnderTouchPoint = null;
+// A variable for the timeout.
```
oferb: Comments should add extra information. In this case, the variable name has the same information, so the comment is unnecessary.

projects/sensory_interface/public/app.js
```diff
@@ -20,29 +28,56 @@ function processData() {
...
     document.getElementById("tableContainer").appendChild(table);
-    addOnClickSoundToTable();
+    addOnClickAndTouchSoundToTable();
 }
 
 function fillRow(values, currentTr) {
     let myValue;
     for (myValue of values) {
-        let td = document.createElement("td");
+        let td = document.createElement("TD");
```
oferb: If `table` and `tr` are lowercase, `td` should be as well, to stay consistent. Is there a reason you changed it?

projects/sensory_interface/public/app.js
```diff
@@ -20,29 +28,56 @@ function processData() {
...
+    startSoundPlayback(valueCalledOn);
+}
+
+function startSoundPlayback(valueCalledOn) {
     if (audioContext.state == "suspended") {
         audioContext.resume();
     }
+    oscillator = audioContext.createOscillator();
+    const MAX_FREQUENCY = 1000;
+    const MIN_FReQUENCY = 100;
```
oferb: `MIN_FReQUENCY->MIN_FREQUENCY`

projects/sensory_interface/public/app.js
```diff
@@ -53,20 +88,47 @@ function startSoundPlayback() {
...
+        if (timeOut != null) {
+            window.clearTimeout(timeOut);
+        }
+    } catch (e) {
+        console.log(e);
+    }
+}
+
+function onCellChange(event) {
+    let changedTouch = event.changedTouches[0];
```
oferb: If we are guaranteed to have at least one element in `changedTouches`, please add it in a comment.
This is to prevent someone else (or, you in the future :) ) considering whether it's a mistake or a case not handled.

projects/sensory_interface/public/app.js
```diff
@@ -20,29 +28,56 @@ function processData() {
...
+        element.addEventListener("touchcancel", stopSoundPlayback);
     }
 }
 
-function startSoundPlayback() {
+function startSoundPlaybackOnClick(event) {
+    currentCellUnderTouchPoint = event.currentTarget;
+    event.preventDefault();
+    stopSoundPlayback(event);
+    let valueCalledOn = event.currentTarget.firstChild.data;
```
oferb: Maybe call `valueCalledOn->selectedValue`?

projects/sensory_interface/public/app.js
mohammad-suliman: I double checked that, it's a lower case. This is strange!! I had a bug because the getTagName returned upper case when called on a td element.
Anyway, I reverted the change.

oferb: Maybe `tagName` is uppercasing everything. In any case, you can do `tagName.toLowerCase()` before you compare, just to be sure.

PR #230, opened 2019-10-08
Sensory intreface project: step 5.

mohammad-suliman:
This PR implements step 5 of the sensory interface project. Note also that it fixes a small bug with the "stopSoundPlayback" function, the bug had no effect on the functionality.

### DIFF
--- a/projects/sensory_interface/public/app.js
+++ b/projects/sensory_interface/public/app.js
@@ -158,10 +158,10 @@ function onRadioChange(radio) {
         minValuePicker.disabled = false;
         maxValuePicker.disabled = false;
     }
-    findMinAndMaxValues();
 }
 
 function findMinAndMaxValues() {
+    // TODO: Add a method to parse the input data to a array of arrays for example, so it can be used here and in processData().
     let input = document.getElementById("dataInput").value;
     let maxValue = -Infinity;
     let minValue = Infinity;
--- a/projects/sensory_interface/public/app.js
+++ b/projects/sensory_interface/public/app.js
@@ -70,9 +70,9 @@ function startSoundPlayback(selectedValue) {
     const MIN_FREQUENCY = 100;
     let minValue = document.getElementById("minValue").value;
     let maxValue = document.getElementById("maxValue").value;
-    maxValue = parseInt(maxValue);
-    minValue = parseInt(minValue);
-    selectedValue = parseInt(selectedValue);
+    maxValue = parseFloat(maxValue);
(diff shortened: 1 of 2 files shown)


projects/sensory_interface/public/app.js
```diff
@@ -131,4 +146,39 @@ function onCellChange(event) {
     let selectedValue = elementUnderTouch.firstChild.data;
     startSoundPlayback(selectedValue);
     event.stopPropagation();
+}
+
+function onRadioChange(radio) {
+    let minValuePicker = document.getElementById("minValue");
```
oferb: If values are already set, don't override them.
Example scenario:
User chooses manual, changes the values, then auto, then wants to go back to manual, but the values were overridden.

projects/sensory_interface/public/app.js
```diff
@@ -131,4 +146,39 @@ function onCellChange(event) {
...
+        maxValuePicker.disabled = true;
+    } else {
+        minValuePicker.disabled = false;
+        maxValuePicker.disabled = false;
+    }
+    findMinAndMaxValues();
+}
+
+function findMinAndMaxValues() {
+    let input = document.getElementById("dataInput").value;
```
oferb: Please add a TODO to have a method to parse the string into a data structure (array of arrays, list of lists etc) and use it both here and in `processData()`.

mohammad-suliman:
Thanks Ofer! A commit will come shortly to address the review comments.
I have a suggestion / consern:
Consider the following use case:
* The user opens the sensory interface project page.
* Then he chooses the "manual" option.
* Then he modifies the "min / max values" for example.
* Here is the issue: he must get back to the "process" button to show the table, the "update" button is still disabled in this stage.

So, do we need to refine this behavior? What do you think?

projects/sensory_interface/public/app.js
```diff
@@ -66,6 +70,15 @@ function startSoundPlayback(selectedValue) {
     const MIN_FREQUENCY = 100;
     let minValue = document.getElementById("minValue").value;
     let maxValue = document.getElementById("maxValue").value;
+    maxValue = parseInt(maxValue);
```
oferb: The value is not necessarily an int. If can be any fractional number.

mohammad-suliman: No problem! Will fix that.

mohammad-suliman:
What about making the "update" button enabled as soon as the user chooses the "manual" option? I think this is a better approach.

PR #231, opened 2019-10-09
sensory interface: step 6.

mohammad-suliman:
This PR implements step 6 of the sensory interface project by updating the URL with the encoded data when the user presses "process", and getting the data from the url if available and inserting  it in the data input field. Note also that the PR fixes a small bug with "findMinAndMaxValues" function.

### DIFF
--- a/projects/sensory_interface/public/app.js
+++ b/projects/sensory_interface/public/app.js
@@ -188,24 +188,21 @@ function findMinAndMaxValues() {
 
 function updateURL() {
     let input = document.getElementById("dataInput").value;
-    let URLDataPart = "data=" + input;
-    URLDataPart = encodeURIComponent(URLDataPart);
+    input = encodeURIComponent(input);
     window.history.pushState({
         state: "examining data"
     },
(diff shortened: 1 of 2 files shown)


projects/sensory_interface/public/app.js
```diff
@@ -182,4 +184,28 @@ function findMinAndMaxValues() {
...
+    URLDataPart = encodeURIComponent(URLDataPart);
+    window.history.pushState({
+        state: "examining data"
+    },
+        "",
+        "/index.html?" + URLDataPart);
+}
+
+function getDataFromURL() {
+    let URL = window.location.href;
```
oferb: Try using `searchParams`: https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams

mohammad-suliman: Thanks so much for this Ofer!! This made the code much prittier.

oferb: Sure thing :))

PR #236, opened 2019-10-15
Add: Anonymous users, suggesters list for each spam message, real time suggestions counter for inbox recycler

cotom972:
1. Anonymous users taken in place when DbMessage instance is created.
You can get the user from: `DbMessage.getFirebaseUser()`

2. Add suggesters list to SmsMessage Proto.

3. Inbox recycler listens to changes in the DB - In case a new spam suggested, it checks whether the user has this spam (By sender & body) in their inbox. If so, it updates the counter.

4. Users can suggest spam only once (This is checked by the ids in the message's suggesters list)
Suggest option will not show up if the user had already suggested.

Overall structure:
FirestoreClient - Writes / Modifies / Deletes message to Firestore.
Any changes that occur are passed to DbMessages.

DbMessages 
Receives the changes from FirestoreClient and updates its list.
Notifies the changes to its listeners - SpamRecyclerAdapter + InboxRecyclerAdapter ( See "3. Inbox recycler ...")


projects/noloan/app/src/main/java/hasadna/noloan/DbMessages.java
```diff
@@ -40,45 +64,74 @@ public void init(List<SmsMessage> messages) {
...
-          smsMessage.toBuilder().setCounter(smsMessage.getCounter() - 1).build();
-      firestoreClient.modifyMessage(smsMessage, newMessage);
-      modifyMessage(newMessage);
-      if (messagesListener != null) {
-        messagesListener.messageModified(messages.indexOf(smsMessage));
+    // Check if user is part of the "suggesters" of this spam message
+    if (smsMessage.getSuggestersList().contains(firebaseUser.getUid())) {
+
+      // Case: Other people had suggested this spam as well, update just the counter (-1)
+      if (smsMessage.getCounter() > 1) {
```
oferb: Why do you need a counter, now that you can use `smsMessage.getSuggestersList().getSize()`?

projects/noloan/app/src/main/java/hasadna/noloan/DbMessages.java
```diff
@@ -99,52 +152,75 @@ public int getIndexById(SmsMessage smsMessage) {
...
       firestoreClient.deleteMessage(smsMessage);
-      if (messagesListener != null) {
-        messagesListener.messageRemoved(index);
-      }
-    } else {
-      Log.e(
-          TAG,
-          "Attempt to remove message by its ID, but message not found\nmessage id: "
-              + smsMessage.getId());
+      notifyListeners(index, smsMessage, Type.REMOVED);
```
oferb: Why did you remove the error logging on not finding a message?

projects/noloan/app/src/main/java/hasadna/noloan/DbMessages.java
```diff
@@ -99,52 +152,75 @@ public int getIndexById(SmsMessage smsMessage) {
...
   }
 
   public void modifyMessage(SmsMessage newMessage) {
-    // Find the existing message in the list - modify
-    if (getIndexById(newMessage) != -1) {
-      messages.set(getIndexById(newMessage), newMessage);
-      if (messagesListener != null) {
-        messagesListener.messageModified(getIndexById(newMessage));
-      }
+    int index = getIndexById(newMessage);
```
oferb: I think you should error log here too.

projects/noloan/app/src/main/java/hasadna/noloan/mainactivity/MainActivity.java
```diff
@@ -49,6 +59,22 @@ protected void onCreate(Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
     setContentView(R.layout.main_activity);
 
+    // Firebase authentication
+    firebaseAuth = FirebaseAuth.getInstance();
```
oferb: I see you have this code twice - here and in `DbMessages`. Try to see if you can avoid that.

projects/noloan/app/src/main/proto/Sms.proto
```diff
@@ -12,6 +12,7 @@ message SmsMessage {
     string id = 4;
     int32  counter = 5;
     bool approved = 6;
+    repeated string suggesters = 7;
```
oferb: Please document that these are Firestore user IDs.

projects/noloan/app/src/main/java/hasadna/noloan/DbMessages.java
cotom972: Thanks, I was wondering about this one.
So perhaps preferably to remove the Proto field `counter`, and just use `SmsMessage.getSuggestersCount()` for this purpose yes?

oferb: Yup.

projects/noloan/app/src/main/java/hasadna/noloan/DbMessages.java
cotom972: :+1:

projects/noloan/app/src/main/java/hasadna/noloan/DbMessages.java
cotom972: Started a branch for adding logs for sensitive scenarios, removed this one now by mistake.
Added back this specific one. I'll push a PR with others later..

projects/noloan/app/src/main/java/hasadna/noloan/mainactivity/MainActivity.java
cotom972: Thanks. It's there from when I first tried to login the user.... 
Removed..

projects/noloan/app/src/main/proto/Sms.proto
cotom972: :+1:

PR #246, opened 2019-11-06
noloan - spliting firestore

gulilo:
… DB spam


projects/noloan/app/src/main/java/hasadna/noloan/InboxRecyclerAdapter.java
```diff
@@ -23,11 +31,47 @@
...
+    Handler handler = new Handler(Looper.getMainLooper());
+
+    // If new spams added to the DB - Check if the spams exists in user's inbox, if so update the
+    // counter of suggesters
+    DbMessages.getInstance().addMessagesListener(new DbMessages.MessagesListener() {
+
+      @Override
+      public void messageAdded(SmsMessage smsMessage) {
+        // Check if user has this message in the inbox
+        int i = messages.indexOf(smsMessage);
```
cotom972: Search inbox messages by Body & Sender..
`SmsMessage` objects coming from the DB, won't probably never be equal to the ones in the inbox<SmsMessage> list.

projects/noloan/app/src/main/java/hasadna/noloan/InboxRecyclerAdapter.java
```diff
@@ -48,8 +49,10 @@ public InboxRecyclerAdapter(ArrayList<SmsMessage> inbox) {
       @Override
       public void messageAdded(SmsMessage smsMessage) {
         // Check if user has this message in the inbox
-        int i = messages.indexOf(smsMessage);
+        int i = DbMessages.findSmsByBody(smsMessage, messages);
         if (i != -1) {
+          messages.remove(i);
+          messages.add(i, smsMessage);
```
cotom972: I think you can use `messages.set(i, smsMessage)` instead?

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/DbMessages.java
```diff
@@ -89,8 +78,8 @@ public void modifyMessage(SmsMessage newMessage) {
   public void addMessagesListener(MessagesListener messagesListener) {
     messagesListeners.add(messagesListener);
   }
-  public void removeMessagesListener(MessagesListener messagesListener)
-  {
+
+  public void removeMessagesListener(MessagesListener messagesListener) {
```
cotom972: IMO this can be removed as it has no usage...
Equivalent to the case where you wouldn't add `Getters/Setters` for all Class members, even if they don't require it.

But perhaps I'm wrong..

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/DbMessages.java
```diff
@@ -123,5 +112,26 @@ public void notifyListeners(int index, SmsMessage smsMessage, Type type) {
     // SmsMessage parameter is passed for cases the index already had been removed from the db list
     void messageRemoved(int index, SmsMessage smsMessage);
   }
+
+  // static function to search a given list
+  public static int findSmsByBody(SmsMessage sms, ArrayList<SmsMessage> list) {
```
cotom972: I think `findSms()` as you proposed before is better :+1: 
Perhaps we can agree as a convention that `findSms()` is by body & sender?

This function finds an SMS by its body AND sender, so `findSmsByBody()` is a little confusing IMO...

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/DbMessages.java
```diff
@@ -123,5 +112,26 @@ public void notifyListeners(int index, SmsMessage smsMessage, Type type) {
...
+
+  public static int findSmsById(SmsMessage sms, ArrayList<SmsMessage> list) {
+    for (int i = 0; i < list.size(); i++) {
+      if (list.get(i).getId().contentEquals(sms.getId())) {
+        return i;
+      }
+    }
+    return -1;
+  }
+
```
cotom972: I think that if it is important for you to keep `DbMessages` a class that handles only DB messages - These functions should be somewhere else, as they were added (if I understand correctly) for the purpose of handling inbox messages, but it's done via the `DbMessages` instance.

So I think for consistency, either:
1. Move them somewhere else - Perhaps `Util.java` ?
2. Switch back to `s/DbMessages/SmsMessages` model, having it for the purpose of handling both Inbox & DB messages. IMO, this solves your question of where to put the `Util.java`'s code as well.
https://github.com/hasadna/monorepo/blob/4580d23fe46c62a38f4b62975c0e86068aa8f1f7/projects/noloan/app/src/main/java/hasadna/noloan/Util.java#L14

BTW - This is the reason that the idea of having `s/DbMessages/SmsMessages` handle both Inbox & DbMessages raised up.

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/FirestoreClient.java
```diff
@@ -32,6 +32,7 @@ public void writeMessage(SmsMessage message) {
   }
 
   public void modifyMessage(SmsMessage oldMessage, SmsMessage newMessage) {
+    Log.d("!!!!!!!!!!!!!!!!!!!!",oldMessage.getApproved() +"    "+newMessage.getApproved());
```
cotom972: Test logs..... ;)

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggetionRecyclerAdapter.java
```diff
@@ -2,42 +2,100 @@
...
+    messages = new ArrayList<>();
+
+    ArrayList<SmsMessage> list = dbMessages.getMessages();
+    for (int i = 0; i < list.size(); i++) {
+      SmsMessage message = list.get(i);
+      if (!message.getApproved()) {
+        Log.d("!!!!!!!!!!!!!!!!!!", "bla");
+        this.messages.add(message);
+      }
+    }
```
cotom972: Perhaps use `foreach` instead? 

    for (SmsMessage sms: dbMessages.getMessages()){
      if(!sms.getApproved()) {
        this.messages.add(sms);
      }
    }

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggetionRecyclerAdapter.java
```diff
@@ -2,42 +2,100 @@
...
+
+import hasadna.noloan.firebase.DbMessages;
+import hasadna.noloan.firebase.FirestoreClient;
 import hasadna.noloan.protobuf.SmsProto.SmsMessage;
 
 public class SuggetionRecyclerAdapter
     extends RecyclerView.Adapter<SuggetionRecyclerAdapter.RecyclerViewHolder> {
 
-  DbMessagesHolder DbMessages;
+  DbMessages dbMessages;
```
cotom972: This is a question I asked myself too...

If DbMessages is a Singleton and the same instance is been used thru out all of the app...

Should it be declared as a Class member for classes that uses it?
Or just be "fetched" with `DbMessages.getInstance()` when ever needed ?

I don't think it actually matters much using it either way, but this thought still raises up...

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggetionRecyclerAdapter.java
```diff
@@ -2,42 +2,100 @@
...
-          public void messageRemoved(int index) {
-            handler.post(() -> notifyItemRemoved(index));
+          public void messageModified(int index) {
+            SmsMessage smsMessage = dbMessages.getMessages().get(index);
+            if (!smsMessage.getApproved()) {
+              int i = messages.indexOf(smsMessage);
+              // already in the list
+              if (i != -1) {
+                messages.remove(i);
+                messages.add(i, smsMessage);
```
cotom972: Same as...
https://github.com/hasadna/monorepo/pull/246#discussion_r343849113

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/DbMessages.java
gulilo: What do you suggest the for a better name?

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/DbMessages.java
gulilo: I don't know where is a better place to put it.
Util is only in the app.

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggetionRecyclerAdapter.java
gulilo: I don't like use foreach, It is less preferment. but done.

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggetionRecyclerAdapter.java
gulilo: You are right, it shouldn't. fixed

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/DbMessages.java
```diff
@@ -114,7 +111,7 @@ public void notifyListeners(int index, SmsMessage smsMessage, Type type) {
   }
 
   // static function to search a given list
-  public static int findSmsByBody(SmsMessage sms, ArrayList<SmsMessage> list) {
+  public static int findSms(SmsMessage sms, ArrayList<SmsMessage> list) {
     for (int i = 0; i < list.size(); i++) {
       if (list.get(i).getBody().contentEquals(sms.getBody())
           && list.get(i).getSender().contentEquals(sms.getSender())) {
```
cotom972: I don't think you need to have it static.
`DbMessages.getInstance()` is static, and returns an object, you can call other functions from.
?

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggetionRecyclerAdapter.java
```diff
@@ -141,7 +130,6 @@ public void bind(SmsMessage sms) {
             FirestoreClient client = new FirestoreClient();
 
             SmsMessage aprroved = sms.toBuilder().setApproved(true).build();
-            Log.d("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!", aprroved.getApproved() + "");
             client.modifyMessage(sms, aprroved);
 
             //client.writeMessage(sms);
```
cotom972: Test comment?

BTW - So in the admin app... We don't use `DbMessages` as our "tool" to pass changes to the database as we do in the NoLoan app?
Meaning... We pass the changes directly via the FirestoreClient? Why is it different?

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/DbMessages.java
```diff
@@ -32,7 +32,7 @@ public static DbMessages getInstance() {
 
   // Search messages by sender & body. Return -1 if no message found
   public int findBySms(SmsMessage smsMessage) {
-    return findSmsByBody(smsMessage, messages);
+    return findSms(smsMessage, messages);
```
cotom972: https://github.com/hasadna/monorepo/pull/246#discussion_r344371555

`s/findBySms/findSms` ?
For consistency, this is confusing IMO...

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SpamRecyclerAdapter.java
```diff
@@ -23,62 +23,54 @@
...
-                handler.post(() -> notifyItemChanged(i));
-              }
-              messages.add(smsMessage);
-              handler.post( () -> notifyItemInserted(messages.size()));
-            }
+      @Override
+      public void messageModified(int index) {
+        SmsMessage smsMessage = dbMessages.getMessages().get(index);
+        if (smsMessage.getApproved()) {
+          int i = messages.indexOf(smsMessage);
```
cotom972: https://github.com/hasadna/monorepo/pull/246#discussion_r343302225

Same goes to here.. 
If a message in the DB have been changed - Its object won't be equal to the "old" one you are trying to change in the local list.

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/DbMessages.java
cotom972: So perhaps `s/DbMessages/SmsMessages`
Have it as a class that holds all functions handling SmsMessages + Holds the DB messages.
And the inbox list stays the same "under" the `InboxRecyclerAdapter`.

>Util is only in the app.

Having this ( https://github.com/hasadna/monorepo/pull/246#discussion_r344373476  ) and needing SmsMessages general functions for the admin app as well (In the admin app you don't use it by the name "inbox" but you use it for other local lists - suggested / approved)

Not having `s/DbMessages/SmsMessages` means you'll need to duplicate some functions to the admin app - I don't think this serves the purpose of "shared modules" you proposed and this whole change.

Not having `s/DbMessages/SmsMessages` and not duplicating them, means you should have `Util` be another shared module as well - IMO this is going a little too far with this idea.

> We can create another class for the inbox but I don't think it is useful.

I agree.

gulilo: >  means you'll need to duplicate some functions to
That is the reason it is static under the DbMessages, So we get access to it from both admin and the app...

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/DbMessages.java
gulilo: But it's not a class function, You don't use any of the class feature with this function.
That way it is static.

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggetionRecyclerAdapter.java
gulilo: Yes. also in the app.
We don't use DbMessages to pass to the DB. It just hold the messages from there.
FirestoreClient is the tool to pass changes to the DB

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/DbMessages.java
gulilo: We have `findSmsById` So I think for consistency this suppose to be `findSmsBy___` . the question is what is this ___..

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/DbMessages.java
cotom972: I still see it as you can get the `DbMessages` instance from both the admin app and the NoLoan app.

But cool, now I understand how you separate things, in terms of functions that aren't used by the `DbMessages` itself, and for external classes usage - static.
:+1:

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/DbMessages.java
cotom972: https://github.com/hasadna/monorepo/pull/246#discussion_r343853004
That's why I said we can agree on `findSms()` as a function that finds by body & sender.
Aside to that for the need of finding a message by its id, have another function: `findSmsById()`

Having all:
`public int findBySms()`
`public int findById()`
`public static int findSms`
`public static int findSmsById()`

For me it sounds confusing and not necessary.
So please just write some "brief" summary in the channel at the end, about the new usage of the functions you changed ok? It'll be easier to follow...

gulilo: The statics functions are for outside use. 
the non-statics function are using the static function on the DB list

and you can search by id or by body and sender

cotom972: Ok, now it's becoming more clear... :+1:

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggetionRecyclerAdapter.java
cotom972: >Yes. also in the app.
We don't use DbMessages to pass to the DB. It just hold the messages from there.
FirestoreClient is the tool to pass changes to the DB

The way the app goes (as it is in the master branch)

All changes concerning the DB are done via the `SmsMessages`, which passes the changes OR not (according to the logic inside the `SmsMessages` functions) to the FirestoreClient.

Please mention that this have been changed in you PR.

projects/noloan/firebase/src/main/java/hasadna/noloan/firebase/DbMessages.java
cotom972: >But it's not a class function, You don't use any of the class feature with this function.

I understand.
So if you're still having non-related DB messages functions under `DbMessages`  - How is that different from the current `SmsMessages`?

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggetionRecyclerAdapter.java
oferb: Hey Gal, the purpose of this PR is to share common code between the main and admin app, but it seems you're also trying to do more stuff here like change the way `DbMessages` work.
If that's the case, please focus this PR just on the narrow goal of sharing code, and any other changes - put in another PR (or several PRs, depending on the change).

projects/noloan/app/src/main/java/hasadna/noloan/Util.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import hasadna.noloan.firebase.DbMessages;
+import hasadna.noloan.firebase.FirebaseAuthentication;
+import hasadna.noloan.firebase.FirestoreClient;
+import hasadna.noloan.protobuf.SmsProto.SmsMessage;
+
+
+// Don't know where to put this code
```
oferb: Please mark comments that should be addressed in the review as `XXX`.
For example:
`// XXX: Don't know where to put this code`

PR #248, opened 2019-11-10
noloan - changed proto to common

gulilo:


projects/noloan/common/src/main/proto/Court.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
```
oferb: What about removing `projects/noloan/common/src/main/proto/Court.proto` and related files?
Does the code use the new files?

projects/noloan/common/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,2 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="hasadna.n.common" />
```
oferb: What is `hasadna.n.common`?

gulilo: It suppose to be `hasadna.noloan.common`
fixing it now

PR #251, opened 2019-11-10
noloan - splitting auth to its own file

gulilo:


projects/noloan/common/src/main/java/hasadna/noloan/common/FirebaseAuthontication.java
```diff
@@ -0,0 +1,43 @@
...
+        user = auth.getCurrentUser();
+      } else {
+        Log.w(TAG, "signInAnonymously:failure", task.getException());
+      }
+    });
+  }
+
+  public boolean containCurrentUserId(List<String> ids) {
+    return ids.contains(user.getUid());
+  }
```
cotom972: This function checks if a user is a suggester of a spam message.

I don't see how it concerns authentication, therefore it doesn't belong to `FirebaseAuthentication`.

You can just check it via the `SmsMessage`:
`SmsMessage.getSuggestersList().contains(FirebaseAuthontication.getInstance().getCurrentUserId());`

OR, if you still think this should be in an external function, have it under `SmsMessages`.

projects/noloan/common/src/main/java/hasadna/noloan/common/FirebaseAuthontication.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.List;
+
+public class FirebaseAuthontication {
+  private static final String TAG = "FirebaseAuthontication";
+
+  private static FirebaseAuthontication instance;
+  private FirebaseUser user;
+  private FirebaseAuth auth;
+
```
cotom972: So this is a Singleton ?

Perhaps mention that in a comment, as you did with the `SmsMessages`.

projects/noloan/common/src/main/java/hasadna/noloan/common/FirebaseAuthontication.java
```diff
@@ -0,0 +1,43 @@
+package hasadna.noloan.common;
+
+import android.util.Log;
+
+import com.google.firebase.auth.FirebaseAuth;
+import com.google.firebase.auth.FirebaseUser;
+
+import java.util.List;
+
+public class FirebaseAuthontication {
```
cotom972: `s/FirebaseAuthontication/FirebaseAuthentication`

projects/noloan/common/src/main/java/hasadna/noloan/common/FirebaseAuthontication.java
gulilo: Don't care.. gone

PR #252, opened 2019-11-10
noloan - changing the way the tabs work in the main app to match the way in th…

gulilo:
…e admin app


projects/noloan/app/src/main/java/hasadna/noloan/mainactivity/MainActivityPagerAdapter.java
```diff
@@ -1,48 +1,44 @@
...
+import java.util.List;
+
 
 public class MainActivityPagerAdapter extends FragmentPagerAdapter {
 
-  private Context context;
+  private List<Fragment> fragments;
+  private List<String> titles;
+
+  public MainActivityPagerAdapter(FragmentManager fm) {
```
oferb: Please use full names: `fm->fragmentManager`

PR #254, opened 2019-11-13
noloan - Fix admin app to use the new proto

gulilo:


projects/noloan/common/src/main/java/hasadna/noloan/common/SmsMessages.java
```diff
@@ -107,7 +107,7 @@ public void undoSuggestion(SmsMessage smsMessage) {
   }
 
   // Search Db/Inbox messages by sender & body. Return -1 if no message found
```
cotom972: Maybe `"// Search SmsMessages list by ... "` ?

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggestionRecyclerAdapter.java
```diff
@@ -2,45 +2,76 @@
...
+
+        if (i != -1 && !smsMessage.getApproved()) {
+          messages.set(i, smsMessage);
+          handler.post(() -> notifyItemChanged(i));
+        } else if (i != -1) {
+          messages.remove(i);
+          handler.post(() -> notifyItemRemoved(i));
+        } else {
+          messages.add(smsMessage);
+          handler.post(() -> notifyItemInserted(messages.size()));
```
cotom972: Please add comments as you've done in `SpamRecyclerAdapter`, describing for which cases these conditions are here.

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggestionRecyclerAdapter.java
```diff
@@ -2,45 +2,76 @@
...
+        }
+      }
+
+      @Override
+      public void messageRemoved(int index, SmsMessage smsMessage) {
+        if (!smsMessage.getApproved()) {
+          int i = SmsMessages.searchMessage(smsMessage, messages);
+          messages.remove(smsMessage);
+          handler.post(() -> notifyItemRemoved(i));
+        }
```
cotom972: If you're already "fetching" the index with `SmsMessages.searchMessage(smsMessage, messages)` which searches by Body & Sender ONLY and returns a result, I'd consider just checking `if(i!=-1)`, and sticking to that index:

`messages.remove(i)`

This can lead to future bugs, as you are removing by Object, and yet notifying by Body & Sender.

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggestionRecyclerAdapter.java
```diff
@@ -2,45 +2,76 @@
...
-    extends RecyclerView.Adapter<SuggetionRecyclerAdapter.RecyclerViewHolder> {
+public class SuggestionRecyclerAdapter
+    extends RecyclerView.Adapter<SuggestionRecyclerAdapter.RecyclerViewHolder> {
+
+  ArrayList<SmsMessage> messages;
 
-  SmsMessages DbMessages;
+  public SuggestionRecyclerAdapter() {
+    messages = new ArrayList<>();
+    SmsMessages dbMessages = SmsMessages.get();
```
cotom972: IMO, this is a little confusing. I think if you need to save it as a local variable, and what you are saving is just the `SmsMessages` instance, as in - you aren't saving the actual dbMessages, then rather name it `smsMessages`

Personally if this local variable is only for the usage of constructing the `SuggestionRecyclerAdapter` and it isn't a class member, I'd just go with `SmsMessages.get()` for this purpose. Not sure how much it matters though, and probably a matter of style.

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SpamRecyclerAdapter.java
```diff
@@ -13,35 +13,66 @@
...
+            SmsMessage smsMessage = dbMessages.getDbMessages().get(index);
+            int i = SmsMessages.searchMessage(smsMessage, messages);
+            if (i != -1 && smsMessage.getApproved()) { // message approved and in the list
+              messages.set(i, smsMessage);
+              handler.post(() -> notifyItemChanged(i));
+            } else if (i != -1) // message not approved but in the list
+            {
+              messages.remove(i);
+              handler.post(() -> notifyItemRemoved(i));
+            } else // message approved but not in the list
```
cotom972: Just to understand...

This is a case which an approved message (written in the DB) had been changed to be disapproved - right?

This can happen (at the moment) only if:
1. This change came from some reason - from the DB
2. Coming from the admin app - Do we have an "undo" for `approve` ?

This cannot come from the NoLoan app right?

projects/noloan/common/src/main/java/hasadna/noloan/common/SmsMessages.java
gulilo: Didn't wrote it, don't care

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggestionRecyclerAdapter.java
gulilo: If it is removed and it wasn't approve it is in my list.
about the `messages.remove(i)` sound good, changed

projects/noloan/common/src/main/java/hasadna/noloan/common/SmsMessages.java
cotom972: Yea I know :)

As part of your change of shared modules, you changed this function to work on general lists. 
I think that it'll be suitable to change the function's description relatively to your change, if you've changed its "goal".

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggestionRecyclerAdapter.java
gulilo: It is used in 2 places so I cached it.
I don't know what it is confusing to you, sorry

gulilo: It is used twice so I cached it.
I am sorry it confusing for you...

projects/noloan/common/src/main/java/hasadna/noloan/common/SmsMessages.java
gulilo: Nope, This is your's code.
The change I did did not approved.

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SuggestionRecyclerAdapter.java
cotom972: Yeah, I got that... A matter of style I suppose....
Cool :+1:

projects/noloan/admin/src/main/java/hasadna/noloan/admin/app/SpamRecyclerAdapter.java
cotom972: Cool, thanks :+1:

projects/noloan/common/src/main/java/hasadna/noloan/common/SmsMessages.java
cotom972: I actually meant the shared modules change. I think this had been approved no?
I'm referring to what's in the master

Anyway, I'm looking at it very simple:

If this function have been changed to be in a shared module for the Admin + NoLoan app
And functions as a function for general SmsMessage now...

I think `"// Search SmsMessages list by ..."`, describes it better.

That's all :)

PR #256, opened 2019-12-06
Change: Spam tab to show only relevant spams for the user

cotom972:
Spam fragment displays only SMS spams that exists in the DB and exists also in the user's inbox.

`SpamRecyclerAdapter` holds a list `inboxSpam` intersected from the spams in the DB, compared by body & sender to the user's inbox messages.

Main title / Spam tab title - Displays number of messages found "similar" (body & sender) in the DB.
(Spam tab shows only messages that are in the inbox)
Inbox title - Number of messages found in the inbox.

### DIFF
--- a/projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
+++ b/projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
@@ -38,20 +38,17 @@ public class SpamRecyclerAdapter
     inboxSpam = new ArrayList<>();
 
     // Fetch spams from DB that are in user's inbox
-    for (int i = 0; i < SmsMessages.get().getInboxMessages().size(); i++) {
-      for (SmsMessage sms : SmsMessages.get().getDbMessages()) {
-        if (SmsMessages.get().getInboxMessages().get(i).getBody().contentEquals(sms.getBody())
-            && SmsMessages.get()
-                .getInboxMessages()
-                .get(i)
-                .getSender()
-                .contentEquals(sms.getSender())) {
-          // Add message from the DB - keep received date of user's message
-          inboxSpam.add(
-              sms.toBuilder()
-                  .setReceivedAt(SmsMessages.get().getInboxMessages().get(i).getReceivedAt())
-                  .build());
-        }
+    for (SmsMessage sms : SmsMessages.get().getInboxMessages()) {
+      int dbIndex = SmsMessages.searchMessage(sms, SmsMessages.get().getDbMessages());
+      if (dbIndex != -1) {
+        // Add message from the DB - keep received date of user's message
+        inboxSpam.add(
+            SmsMessages.get()
+                .getDbMessages()
+                .get(dbIndex)
+                .toBuilder()
+                .setReceivedAt(sms.getReceivedAt())
+                .build());
       }
     }
 
@@ -82,6 +79,7 @@ public class SpamRecyclerAdapter
                     SmsMessages.get()
                         .searchMessage(SmsMessages.get().getDbMessages().get(index), inboxSpam);
                 if (inboxSpamIndex != -1) {
+                  // Modify message - keep received date of user's message
                   inboxSpam.set(
                       inboxSpamIndex,
                       SmsMessages.get()
--- a/projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
+++ b/projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
@@ -31,7 +31,7 @@ public class SpamRecyclerAdapter
     extends RecyclerView.Adapter<SpamRecyclerAdapter.RecyclerViewHolder> {
   private static final String TAG = "SpamRecyclerAdapter";
 
-  List<SmsMessage> inboxSpam;
+  ArrayList<SmsMessage> inboxSpam;
 
   public SpamRecyclerAdapter() {
     Handler handler = new Handler(Looper.myLooper());
--- a/projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
+++ b/projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
@@ -16,7 +16,9 @@ import android.widget.TextView;
 
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Calendar;
+import java.util.List;
 import java.util.Locale;
 
 import hasadna.noloan.common.FirebaseAuthentication;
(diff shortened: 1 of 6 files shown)


projects/noloan/app/src/main/java/hasadna/noloan/mainactivity/MainActivity.java
```diff
@@ -162,17 +164,24 @@ public boolean onNavigationItemSelected(MenuItem item) {
...
+    statusTitle.setText(String.valueOf(SmsMessages.get().countInboxSpam()));
 
-    // Tab's titles
+    // Number of messages in the inbox
     tabLayout
         .getTabAt(0)
         .setText(
             getString(R.string.inboxFragment_title, SmsMessages.get().getInboxMessages().size()));
+
+    // Number of messages that the user had suggested
```
cotom972: To be fixed next push. This isn't correct. Suppose to be number of messages relevant to the user's inbox.

projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
```diff
@@ -29,26 +31,80 @@
...
+  ArrayList<SmsMessage> inboxSpam;
+
   public SpamRecyclerAdapter() {
     Handler handler = new Handler(Looper.myLooper());
+    inboxSpam = new ArrayList<>();
+
+    // Fetch spams from DB that are in user's inbox
+    for (int i = 0; i < SmsMessages.get().getInboxMessages().size(); i++) {
+      for (SmsMessage sms : SmsMessages.get().getDbMessages()) {
+        if (SmsMessages.get().getInboxMessages().get(i).getBody().contentEquals(sms.getBody())
```
gulilo: Why didn't you use 'searchMessage'?

projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
```diff
@@ -29,26 +31,80 @@
...
+    // Keep the received date of the user's message.
     SmsMessages.get()
         .setMessagesListener(
             new SmsMessages.MessagesListener() {
               @Override
               public void messageAdded(SmsMessage newMessage) {
-                handler.post(() -> notifyItemInserted(SmsMessages.get().getDbMessages().size()));
+                int inboxIndex =
+                    SmsMessages.get()
+                        .searchMessage(newMessage, SmsMessages.get().getInboxMessages());
```
gulilo: Cache the 'InboxMessages' since it is used in multiple places.
Same for 'DbMessages'.

projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
```diff
@@ -29,26 +31,80 @@
...
+                                  .getReceivedAt())
+                          .build());
+                  handler.post(() -> notifyItemChanged(inboxSpamIndex));
+                }
               }
 
               @Override
               public void messageRemoved(int index, SmsMessage smsMessage) {
-                handler.post(() -> notifyItemRemoved(index));
+                int inboxSpamIndex = SmsMessages.get().searchMessage(smsMessage, inboxSpam);
```
gulilo: It can be call index, since there is not other indexes around.
Same in modified.

projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
```diff
@@ -29,26 +31,80 @@
...
+                }
               }
 
               @Override
               public void messageModified(int index) {
-                handler.post(() -> notifyItemChanged(index));
+                int inboxSpamIndex =
+                    SmsMessages.get()
+                        .searchMessage(SmsMessages.get().getDbMessages().get(index), inboxSpam);
+                if (inboxSpamIndex != -1) {
```
gulilo: What are you doing here? having hard time following

projects/noloan/app/src/main/java/hasadna/noloan/mainactivity/SpamFragment.java
```diff
@@ -14,13 +14,14 @@
 import noloan.R;
 
 public class SpamFragment extends Fragment {
+  static final String TAG = "SpamFragment";
 
   private OnFragmentInteractionListener fragmentInteractionListener;
   private RecyclerView recyclerView;
-  private SpamRecyclerAdapter spamRecyclerAdapter;
+  SpamRecyclerAdapter spamRecyclerAdapter;
```
gulilo: Is it used somewhere?

projects/noloan/common/src/main/java/hasadna/noloan/common/SmsMessages.java
```diff
@@ -42,7 +42,7 @@ public void suggestMessage(SmsMessage smsMessage) {
     // 1. Add user as a "suggester"
     int index = searchDbMessage(smsMessage);
     if ((index != -1)
-        && dbMessages
+        && !dbMessages
```
gulilo: Why this change?

projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
cotom972: Good point, forgot about it. Thanks
Fixed :+1:

projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
cotom972: I got the "hang" of what you were saying about having some `SmsMessages` function statics, and that it's easier to use (instead of `.get()` each time) as I wrote this...

I didn't want to push these changes (concerning SmsMessages static functions) in this PR, so left it like that for now.
Lets raise this idea again in the channel...

projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
cotom972: It looks a little "messy" to me too, that's why I added the comment above all of these functions. (Add / Modify / Remove)
>// Listen to db smsMessages. Update just relevant messages that are in the inbox as well.
>// Keep the received date of the user's message.

I'll add another comment for this function specifically...

projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
cotom972: `index` is part of the function's parameter. 

`inboxSpamIndex` feels easier and faster to "grasp" to which list this index belongs to IMO.

projects/noloan/app/src/main/java/hasadna/noloan/mainactivity/SpamFragment.java
cotom972: MainActivity - `updateTitles()`
Overall I think updating the titles can be done in a better way than it is at the moment, but for a later task...

projects/noloan/common/src/main/java/hasadna/noloan/common/SmsMessages.java
cotom972: It's a bug fix. I think you accidentally removed the `!` here: 
https://github.com/hasadna/monorepo/commit/a14fceba8dc579302d1342bcfb0e83195a269b21#diff-e2258fc75a728aa3ae8880a986424ae0L61

It prevents a user from suggesting a spam more than once.

projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
gulilo: > index is part of the function's parameter.

Right...
I prefer shorter names so it bother me.

projects/noloan/app/src/main/java/hasadna/noloan/mainactivity/SpamFragment.java
gulilo: why main activity is the responsible to update the title?
maybe add a update title in each fragment and main activity just call it, so mainActivity don't need to know about the recycler.

projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
gulilo: getInboxMessages and getDbMessages cannot be static they use the class fields.

just before the call for setMessagesListener (bad name BTW, you are adding not setting) save thous into a variable and use it instead.
calling the same function more then once is not a good thing, It is maybe OK in our case because it is just a get function, but generally it is something you want to avoid.

Plus it make the code longer then it is actually need to be and harder to read.

projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
gulilo: maybe with the cached list it will be easier to read.
Plus the way it written
as 
a 
single
thing
every
line
kinda make it hard to read. I think ff doing it and it is annoying...

projects/noloan/common/src/main/java/hasadna/noloan/common/SmsMessages.java
gulilo: Opss?

projects/noloan/app/src/main/java/hasadna/noloan/SpamRecyclerAdapter.java
```diff
@@ -31,24 +31,22 @@
     extends RecyclerView.Adapter<SpamRecyclerAdapter.RecyclerViewHolder> {
   private static final String TAG = "SpamRecyclerAdapter";
 
-  ArrayList<SmsMessage> inboxSpam;
+  List<SmsMessage> inboxSpam;
```
gulilo: Why the change from arrayList to List?

